PR #22, opened 2018-10-10
[OCA]12567 Location structure needs to reflect hierarchy of Buildings

smangukiya:
Modified module:
> field-service

### DIFF
--- a/fieldservice/README.rst
+++ b/fieldservice/README.rst
@@ -85,6 +85,7 @@ Contributors
 * Wolfgang Hall <whall@opensourceintegrators.com>
 * Maxime Chambreuil <mchambreuil@opensourceintegrators.com>
 * Steve Campbell <scampbell@opensourceintegrators.com>
+* Sandip Mangukiya <smangukiya@opensourceintegrators.com>
 
 Other credits
 ~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


dreispt:
We're building a bloated monolithic module here.
IMO this should be optional, and so separated into it's won addon module.

etobella:
That was also my comment on #23

wolfhall:
I can see that with specific workflows, a building hierarchy is not needed.  However, I had several customers who needed more details.  when I worked with UPS and FedEx and large Utility companies one of the key problem was that the delivery or service point is often not the address point but a different location.  The good thing about the model is that all the fields are optional.  It would be great to have a configuration setting to turn some detailed location fields on/off.

JordiBForgeFlow:
In my past experience in service management / plant maintenance in SAP this would be done using a hierarchical structure,  like the concept of Location in stock. In SAP they call it "Technical Location". And you can define a hierarchical tree that maps to physical places. The location does not necessarily need to be within your company.

IMHO adding so many fields as in this PR is never going to solve all the problems, whereas adding the field "Technical Location" would do.

wolfhall:
I remember (a little fuzzy) using both functional and technical location for elevator equipment location representation.  If I understand your suggestion you would like a generic, flexible hierarchy that allows us to define any "technical location" and tie it to a parent record (either another technical location or a functional location.   Max and I are working with a customer right now who needs the details - we can see if this makes sense.  I like the suggestion.

dreispt:
A Location hierarchy is important. But that’s not what’s happening here.
And there is a chance that the customer_id field is a duplication of the partner_id field.

max3903:
Please review #27

Issue #4, opened 2018-09-10
Manage Vehicles


max3903:
- [x] Add "Manage Vehicles" in the settings
- [x] Add Vehicles menu item in Master Data
- [x] Add Vehicle to the Route
- [x] Add Vehicles menu item in Reporting

brian10048:
This will require adding a dependency to fleet in order to add fields to the fleet vehicle model.  ex: fsm_vehicle boolean that could be used to filter the Vehicles in the Master Data > Vehicles menu item view action 

Should the "Manage Vehicles" setting install a 'fieldservice_fleet' add-on that has the fleet dependency and accomplishes remaining/future items in the list?

wolfhall:
We are treating FSM vehicles as a separate entity from fleet vehicles.  Initially we thought we could re-use the fleet vehicle tables but there is actually a different meaning between the two:

Fleet Vehicles in the standard module are assets that require maintenance and are tracked for mileage purposes.  Most dispatchers don't concern themselves with these aspects - they don't need classic fleet management functions.

Field Service Vehicles are a tool used in FSM that serves as an inventory location, has additional attributes (such as: has forklift, side load, ladder, etc.) and can be tied to a fleet service vehicle when needed.   keeping it separate allows us to be more flexible and not require the Fleet Management Module.  We are also dealing with many companies who use field service contractors who own their own vehicles.

max3903:
Hello @brian10048,

The idea here is to start with a fsm.vehicle object.

If "Manage Vehicles" is selected in the settings, the "Install fleet module" (not sure about the label yet) option would appear. Checking this box would install the fieldservice_fleet module. In this module, the fsm.vehicle object would inherit the vehicle object provided by the fleet module. Like fsm.person inherits res.partner.

PR #23, opened 2018-10-11
[ADD] Vehicle

brian10048:
Is this what is needed for #4 ?

Still need to add some reporting features

### DIFF
--- a/fieldservice/__manifest__.py
+++ b/fieldservice/__manifest__.py
@@ -25,6 +25,7 @@
         'views/res_partner.xml',
         'views/fsm_location.xml',
         'views/fsm_person.xml',
+        'views/fsm_vehicle.xml',
         'views/fsm_order.xml',
         'views/fsm_route.xml',
         'views/menu.xml',
--- a/fieldservice/models/__init__.py
+++ b/fieldservice/models/__init__.py
@@ -8,6 +8,7 @@ from . import (
     fsm_tag,
     fsm_location,
     fsm_person,
+    fsm_vehicle,
     fsm_order,
     fsm_route,
 )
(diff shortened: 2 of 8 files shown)


max3903:
Thanks @brian10048. Nice work!

Let me know if you need help signing the CLA and fixing Travis errors.

brian10048:
Glad to be a part @max3903 

I emailed my signed CLA with GitHub login yesterday once the comment was posted here. 

Looks like Travis errors are good now as well!

max3903:
@brian10048 I processed your CLA. :+1:

etobella:
I am probably late, but why are you adding vehicles on the main module? IMHO, it should be made modular. So if a company has vehicles, it should be added as a new module. If all is added on the same module, you may have all functionality but in some cases only 30% is used

max3903:
The future module `fieldservice_vehicle` is meant to depend on `fleet` to extend the fsm.vehicle with fleet features.

The idea is to provide the basic features with settings in a single module, and advanced ones with additional modules.

etobella:
Vehicles should not be a basic feature. On your documents, you expected to add features for even doctors, but they usually don't use  vehicles. I think the main module should have the service order and all the shared models. All the other models should be added on new modules.
If you add all the functionalities on the initial module, it is harder to use it, and some possible may loose the interest on it.

max3903:
Vehicle feature is only visible if you activate it in the settings, like invoicing vs accounting or product template vs variants.

brian10048:
The vehicle features are hidden by default in the core module and require
no additional dependencies

The nature of field service would most of the time require that you use a
vehicle to travel to the location being serviced. If one doesn't need to
track vehicle data, the feature can remain off.

etobella:
Yes, I know, but you are adding unnecessary features on the initial module. That is what i mean.
For example, when bank-payment was created, all could be added on a single module, but it was splitted in order to not include too much functionality on a single module. I think it is the best approach, but it is just my opinion.

PR #27, opened 2018-10-18
[ADD] location, territory, branch, district, region

max3903:
Fixing comments on #22


fieldservice/models/branch.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Branch(models.Model):
+    _name = 'branch'
+    _description = 'branch'
```
dreispt, line 9: I think this overlaps with `hr.department`.
That's what I'm using today for the same purpose.

fieldservice/models/branch.py
```diff
@@ -0,0 +1,12 @@
...
+
+    name = fields.Char(string='Name')
+    district_id = fields.Many2one('district', string='District')
```
dreispt, line 12: If we add a `partner_id` Address field to the Branch/Department, we automatically have a full address.
If we need additional fields for it, modules like `base_location` would do that for us.

fieldservice/models/district.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo import fields, models
+
+
+class District(models.Model):
+    _name = 'district'
+    _description = 'District'
+
+    name = fields.Char(string='Name')
+    region_id = fields.Many2one('region', string='Region')
```
dreispt, line 12: Instead of the fixed District -> Region -> ... multimodel fixed structure, we could have a flexible single model hierarchy structure. We could even include "Branch" and "Territory" as hierarchy layers, and avoid forcing a dependency on HR (it would then become an optional link).

fieldservice/models/fsm_location.py
```diff
@@ -9,6 +11,13 @@ class FSMLocation(models.Model):
     _inherits = {'res.partner': 'partner_id'}
```
dreispt: New style delegation inheritance: remove _inherits and instead add delegage=True on the partner_id field.

fieldservice/models/fsm_location.py
```diff
@@ -21,18 +30,27 @@ class FSMLocation(models.Model):
...
-    branch = fields.Char(string='Branch', size=35)
-    district = fields.Char(string='District', size=35)
-    region = fields.Char(string='Region', size=35)
-    timezone = fields.Char(string='Timezone', size=35)
+    location_id = fields.Many2one('location', string='Location')
+    territory_id = fields.Many2one('territory', string='Territory')
+    branch_id = fields.Many2one('branch', string='Branch')
+    district_id = fields.Many2one('district', string='District')
+    region_id = fields.Many2one('region', string='Region')
+    timezone = fields.Selection(_tz_get, string='Timezone')
```
dreispt, line 38: Yes.

fieldservice/models/location.py
```diff
@@ -0,0 +1,32 @@
...
+    _description = 'Location'
+    _parent_name = "parent_id"
+    _parent_store = True
+    _parent_order = 'name'
+    _order = 'parent_left'
+    _rec_name = 'complete_name'
+
+    name = fields.Char(string='Name')
+    complete_name = fields.Char("Full Location Name",
+                                compute='_compute_complete_name', store=True)
```
dreispt: +index=True

fieldservice/models/location.py
```diff
@@ -0,0 +1,32 @@
...
+    parent_id = fields.Many2one('location', string='Parent Location',
+                                ondelete='restrict')
+    notes = fields.Html(string='Notes')
+    parent_left = fields.Integer('Left Parent', index=True)
+    parent_right = fields.Integer('Right Parent', index=True)
```
dreispt, line 24: I think these are not used anymore, but I need to check (maybe on v12 only)

fieldservice/models/fsm_location.py
```diff
@@ -21,18 +30,27 @@ class FSMLocation(models.Model):
                                   auto_join=True)
     tag_ids = fields.Many2many('fsm.tag',
                                string='Tags')
-    building = fields.Char(string='Building', size=35)
-    floor = fields.Char(string='Floor', size=35)
-    unit = fields.Char(string='Unit', size=35)
-    room = fields.Char(string='Room', size=35)
```
dreispt, line 27: These building/floor/room details don't need to be structured data.
When needed they can perfectly be included in the Location's name or Description.

dreispt (commented):
I suggest model design improvements.

fieldservice/models/branch.py
max3903: @dreispt The idea behind branch is mostly where the company has offices. Department is not necessarily geographically centralized...

fieldservice/models/district.py
max3903: @dreispt Your comments makes total sense from a functional perspective, but I think @wolfhall has specific geospatial reasons behind those models.

@wolfhall Am I right?

fieldservice/models/location.py
max3903: @dreispt v12 ;)

fieldservice/models/fsm_location.py
max3903: @dreispt that's why we removed them.

fieldservice/models/location.py
max3903: ok

fieldservice/models/fsm_location.py
max3903: ok

osi-scampbell (approved):

osi-scampbell (approved):

osi-scampbell (approved):

wolfhall (approved):
Looks good

dreispt:
Can we have the Branch /Territory hierarchy instead?

brian10048:
I like where this is going, but from reading the code and from a user's perspective the location model is confusingly named when we have already established the fsm_location model.  As @wolfhall suggested in #22  the newer location model aims to represent a specific point of service within the fsm_location, so could we rename it service_point or something similar?

That being said, wouldn't we have a relationship of one fsm_location to many service_points?  Each service_point is unique to that fsm_location.  The configuration area could establish multiple service_point_types that create the hierarchy of service_points within the fsm_location others have mentioned needing.   

~~Let me know if this makes sense and I'll put something in motion to demonstrate~~ Please see #33

Issue #29, opened 2018-10-31
Concerns about compatibility/overlap with Maintenance modules


dreispt:
I'm sorry for not bringing this up earlier:

Field service includes maintenance, but has a broader scope than that.
But looking at the standard "[maintenance](https://github.com/odoo/odoo/tree/12.0/addons/maintenance)" app, I see some common ground with FS Orders.

I wonder if it is not a better design to have FS Orders based on Maintenance requests, so that we can leverage addons (and avoid duplicating) [maintenance community features](https://github.com/OCA/maintenance/tree/10.0).

dreispt:
![image](https://user-images.githubusercontent.com/1246629/47788775-9db38700-dd0a-11e8-989a-5633ed1a5e8b.png)

![image](https://user-images.githubusercontent.com/1246629/47788791-aa37df80-dd0a-11e8-9311-8c804985f6c4.png)

![image](https://user-images.githubusercontent.com/1246629/47788819-bfad0980-dd0a-11e8-97ac-5ef938152d08.png)

![image](https://user-images.githubusercontent.com/1246629/47788840-cfc4e900-dd0a-11e8-9327-8330782238bd.png)

![image](https://user-images.githubusercontent.com/1246629/47788880-e8cd9a00-dd0a-11e8-8135-57b8f7ceb137.png)

dreispt:
Proof of concept: <s>#30</s>#31
Only focuses on the Orders model.
Needs work on Views and Security Groups.

max3903:
@dreispt Can we close this one?

PR #31, opened 2018-10-31
[REF] FSM Order based on maintenance.request

dreispt:
Proof-of-concept for #29

### DIFF
--- a/fieldservice/__manifest__.py
+++ b/fieldservice/__manifest__.py
@@ -9,8 +9,8 @@
     'author': 'Open Source Integrators, Odoo Community Association (OCA)',
     'website': 'https://github.com/OCA/field-service',
     'depends': [
-        'base_geolocalize',
-        'mail',
+        # 'base_geolocalize',
+        'maintenance',
         'web_timeline',
     ],
     'data': [
--- a/fieldservice/data/fsm_stage.xml
+++ b/fieldservice/data/fsm_stage.xml
@@ -1,51 +1,51 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
-    <record id="fsm_stage_new" model="fsm.stage">
+    <record id="fsm_stage_new" model="maintenance.stage">
         <field name="name">New</field>
         <field name="sequence">10</field>
     </record>
 
-    <record id="fsm_stage_confirmed" model="fsm.stage">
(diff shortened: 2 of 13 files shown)


max3903:
@dreispt I understand this is a proof of concept but eventually you will want this to be merged. Can you do your POC as a separate module?

dreispt:
This changes the FSM core Models, I don't see how I can make it in a separate model.
In the meanwhile I made a 3rd commit, making the Dashboard and Order views work, should be available on Runbot soon.

dreispt:
💚

dreispt:
Can we go ahead with this approach?

max3903:
As a separate module, yes.

Not all field service activities require maintenance features.

Please review the list of use cases.

dreispt:
I don't see how this can be a separate module.

The point is that `fieldservice` is reimplementing features that already exist on `maintenance`.
The goal is for `fieldservice` to to override and expand `maintenance`, to encompass the more generic concept of Field Service. So it is foundational.

My PoC intended to show that this can be done with little impact for the FSM roadmap.

rvalyi:
Or else we have to extract a common basis module that would be used both for FSM and maintenance, thus you could still have FSM without the full blown maintenance dependency. Not sure how practical it is.

dreispt:
@rvalyi maintenance is an upstream Odoo core module, so we can’t change it.

From a practical POV I don’t see an alternative to the implementation in this PR. But I may be missing something.

dreispt:
The issue here is purely undervote hood technical. From a user PoV there is little difference:
I challenge you to visit Runbot for this branch and for main 11.0 and compare (please ignore the to disappear Maintenance menu).

brian10048:
I tried it on the runbot and yes there is no major difference. However the order is the main transaction type of model in field service, so shouldn't it be its own entity? Much like a maintenance request is for the Maintenance module, there are tasks in Projects, or work orders in MRP, we need fsm orders for FSM.

Using a maintenance request to replace fsm orders only satisfies some use cases. For example, the Construction industry use cases would be more suited to following the flow of mrp work orders or maybe a project task. The FSM order is used to organize and optimize the scheduling of work to be done at remote locations. Once on site, the work being done might follow the workflow of a maintenance request at times, a task in an ongoing project, or an mrp order depending on the individual configuration.

max3903:
Please have a look at #66

Issue #11, opened 2018-09-10
fieldservice_stock


max3903:
License: AGPL-3

# Module description
Odoo Field Service App allows to manage service requests and orders. The app does not support the logistic side of those activities: Sometimes, the dispatcher need to ship materials to the field service person or to the serviced location.

This module allows the dispatcher to request materials to be shipped to the customer location or serviced location and follow up with the Inventory team.

# Usage
* Create a service order with a customer
* In Materials, add products with quantity
* Click on Confirm
  * The delivery order is created
  * The transfer is set on the ticket
  * The state is updated to Confirmed
  * The product, description and quantity requested cannot be changed
* Transfer partially the delivery order
  * The state is updated to Partially Shipped
* Transfer completely the delivery order
  * The state is updated to Done

# 0.0.1

## Dependencies
* fieldservice
* stock

## Configuration
- [x] Add the "Use Odoo Logistics features" option in the setting panel
- [x] Add a view location "Field"
- [x] Add a view location "Vehicles" under the main warehouse
- [x] Add stock operations: "Output to Vehicle", "Vehicle to Location", "Location to Vehicle" and "Vehicle to Input"

## Master Data
- [x] Add a many2one field "Inventory Location" to fsm.location. Values are filtered with the stock location type: internal, customer or vendor.
- [x] Add a many2one field "Inventory Location" to fsm.vehicle. Values are filtered with the stock location type: internal. The field is only visible by users in the group "Manage Vehicles"

# 0.0.2

## Operations
FSM Order Line
- [x] Add a fsm.order.line object
- [x] Add a required char field "Description" (name).
- [x] Add a many2one field "Product" (product_id) to product.product. Only stockable product can be selected. Like the sale order line, selecting the product set the description.
- [x] Add a float field "Quantity Requested" (qty_ordered)
- [x] Add a readonly float field "Delivered" (qty_delivered)
- [x] Add a selection field "State" (state) with: Draft, Confirmed, Partially Shipped, Done. If the state is not Draft, the other fields are readonly.
- [ ] ~~Add a readonly many2one field "Transfer" (picking_id) to stock.picking~~

FSM Order
- [x] Add a one2many field "Materials" (line_ids) to fsm.order.line
- [x] Add a button to confirm all draft lines. Like a sales order, the confirmation creates a procurement in the location of the customer and relate the delivery order to the ticket (picking_id is set)
- [x] When the delivery order is transferred, the quantity delivered is updated.
[...]

PR #38, opened 2018-11-24
[ADD] fieldservice_stock 0.0.1

brian10048:
This should satisfy the tasks under 0.0.1 in #11.  

Open to suggestions on better ways to setup the stock operation types


max3903 (approved):

max3903:
@brian10048 Thank you!

Issue #37, opened 2018-11-23
fieldservice 0.0.2


max3903:
# Configuration
## Field Service Person Category
- [x] Add a menu item in Configuration > Persons
- [x] Add a fsm.person.category object
- [x] Add a required char field "Name"

## Field Service Order Template
- [x] Add a menu item Templates in Configuration > Orders
- [x] Add a fsm.order.template object
- [x] Add a required char field "Name"
- [x] Add a many2many field "Categories" (category_ids) to fsm.person.category

## Territory
- [x] Add a many2many fields "Field Service Persons" (fsm_person_ids) to fsm.person

# Master Data
## Field Service Person
- [x] Add a many2many field "Categories" (category_ids) to fsm.person.category
- [x] Add a many2one field "Working Schedule" (calendar_id) to resource.calendar

## Field Service Location
- [x] Add a many2one field "Preferred Field Service Person" to fsm.person.
- [x] ~~Setting the Preferred FS Person adds the person to the list of persons that serves the related territory~~

# Operations
## Orders
- [x] Add a many2many field "Categories" (category_ids) to fsm.person.category
- [x] Add a many2one field "Template" (template_id) to fsm.order.template
- [x] Selecting the template fills in the categories
- [x] ~~Add an onchange to set the FS person automatically using the preferred FS person from the location~~
- [x] ~~Filter the list of FS persons with the ones that serves the territory of the location~~

Issue #9, opened 2018-09-10
Manage Skills: fieldservice_skill


max3903:
# Module description

Odoo Field Service App allows to assign any FS person to any FS order. For some organizations, they cannot assign any FS person to any order. They needs to manage the skills of the FS persons and reduce the list of potential FS persons on an order to avoid wrong assignment and mis-configuration. The skills of the selectable FS persons on a service order needs to match the required skills to perform the order.

This modules allows you to set skills on categories to automatically set those skills on the order. It also allows you to set the skills to the field service person with their level. On the order, the list of field service persons is filtered with the skills and the location so you can only assign the right person to the order.

# Configuration
* Create or edit your categories to set skills
* Create or edit your persons to set their skills and skill levels
* Create or edit your territories to set their field service persons
* Create or edit your locations to set their territories

# Usage
* Create an order, set the template and the location. 
  * The categories and skills are set based on the template.
  * The field service person is set to the preferred person if skills matches.
  * The list of field service persons is filtered with the one serving the location (using the territory) and having the skills

# Dependencies
- [x] #37 
- [x] hr_skill 11.0.2.0.0: https://github.com/OCA/hr/issues/503

# Configuration
## Settings
- [x] Add the "Manage Skills" option in the setting panel to install the fieldservice_skill module
## Categories
- [x] Add a many2many field "Skills" (skill_ids) to hr.skill on fsm.person.category
## Skills
- [x] Add a Skills menu item
- [x] Add a fsm.person.skill object
- [x] Add a required many2one field "Field Service Person" (fsm_person_id) to fsm.person
- [x] Add a required many2one field "Skill" (skill_id) to hr.skill
- [x] Add a required selection field "Skill Level" (skill_level_id) with: 0 Junior, 1 Intermediate, 2 Senior, 3 Expert

# Master Data
## Persons
- [x] Add a one2many field "Skills" (skill_ids) to fsm.person.skill using fsm_person_id

# Operations
## Orders
- [x] Add a many2many field "Skills" (skill_ids) to hr.skill
- [x] Selecting the template fills in the categories and skills ~~and filter the list of FS persons~~
- [x] ~~Set the FS person automatically using the preferred FS person from the location, if the skills match~~

# Reporting
## Skills
- [x] Add a menu item "Skills" in Reporting for fsm.person.skill
- [x] Add a pivot view for fsm.person.skill

dreispt:
[...]

PR #43, opened 2018-12-04
[ADD] fieldservice_skill

osi-scampbell:
[ADD] Separated Fieldservice_skill into separate module

#9 

Depends on:
- [x] #37 
- [x] https://github.com/OCA/hr/issues/503


max3903:
@osi-scampbell We are still missing 2 features from the issue #9 : 

- [x] Selecting the template fills in the categories and skills ~~and is used to set the default filter on the list of FS persons~~
- [x] ~~Set the FS person automatically using the preferred FS person from the location, if the skills match~~

PR #39, opened 2018-11-26
[IMP] fieldservice 0.0.2

brian10048:
Closes #37

PR #51, opened 2018-12-12
[IMP] Add working schedule on the fsm.person

max3903:
#37


max3903:
Superseeds #39

brian10048 (commented):
I think this will require a dependency to resource?

max3903:
@brian10048 Good catch, Thanks! Fixed in #52

PR #68, opened 2018-12-19
Scampbell phone scope

osi-scampbell:
Added scope and phone to helpdesk


pedrobaeza:
`helpdesk` is the enterprise one? @max3903 you know what happens with enterprise modules...

Issue #65, opened 2018-12-18
Field Service Stock 11.0.0.3.0


max3903:
# fsm.equipment
- [x] Add a computed field "Current Inventory Location" (current_stock_location_id) to show the current inventory location of the related serial number

# fsm.order.return
- [x] Add an object "Return Line" fsm.order.return
- [x] Add a char field "Description" (name) to fsm.order.return
- [x] Add a many2one field "Product" (product_id) on fsm.order.return to product.product
- [x] Add a float field "Quantity Returned" (qty_returned)
- [x] Add a float field "Quantity Received" (qty_received)
- [x] Add a many2one field "Order" (order_id) to fsm.order

# fsm.order
- [x] Add a one2many field "Return Lines" (return_ids) to fsm.order.return using order_id
- [x] Add a smart button to display the number and the list of receipts related to the order
- [x] When the order is completed, the receipt is created

PR #81, opened 2018-12-31
[imp] fieldservice stock

smangukiya:
ref: https://github.com/OCA/field-service/issues/65

@max3903 : Please review it. Thanks

### DIFF
--- a/fieldservice_stock/models/fsm_equipment.py
+++ b/fieldservice_stock/models/fsm_equipment.py
@@ -1,0 +1,28 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class FSMEquipment(models.Model):
+    _inherit = 'fsm.equipment'
+    _description = "FSM Equipment Inherits"
+
+    product_id = fields.Many2one(
+        'product.product', string='Product', required=True)
+    lot_id = fields.Many2one(
+        'stock.production.lot', string='Serial #', required=True)
+    current_stock_location_id = fields.Many2one(
+        'stock.location', string='Current Inventory Location',
+        compute='_compute_current_stock_loc_id')
+
+    @api.multi
+    def _compute_current_stock_loc_id(self):
+        for equipment in self:
+            quants = self.env['stock.quant'].search(
+                [('lot_id', '=', self.lot_id.id)], order="id desc")
+            if quants:
+                equipment.current_stock_location_id = \
+                    quants[0].location_id.id or False
+            else:
+                equipment.current_stock_location_id = False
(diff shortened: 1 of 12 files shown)


fieldservice_stock/models/fsm_equipment.py
```diff
@@ -0,0 +1,23 @@
...
+    current_stock_location_id = fields.Many2one(
+        'stock.location', string='Current Inventory Location',
+        compute='_get_current_stock_loc_id')
+
+    @api.multi
+    def _get_current_stock_loc_id(self):
+        for equipment in self:
+            if equipment.lot_id and equipment.lot_id.product_id:
+                equipment.current_stock_location_id = \
+                    equipment.lot_id.product_id.property_stock_inventory \
```
max3903: @smangukiya `current_location_location_id` should be set to the current location of the serial number: when the serial number is delivered to the customer location, the current location should change to `Partner Location / Customers`.

fieldservice_stock/models/fsm_order.py
```diff
@@ -35,15 +35,27 @@ def _default_warehouse_id(self):
...
+                 picking.picking_type_id.code == 'outgoing'])
+            order.return_count = len(
+                [picking for picking in order.picking_ids if
+                 picking.picking_type_id.code == 'incoming'])
 
     def action_confirm(self):
         if self.location_id and self.line_ids and self.warehouse_id:
             self.line_ids._confirm_picking()
+            if self.return_ids:
+                self.return_ids._confirm_picking()
```
max3903: @smangukiya It doesn't work if I have a return line (inbound) and no line (outbound). I keep getting the error message below. Confirming should confirm the draft lines and draft returns if there is any.

max3903 (changes requested):

fieldservice_stock/models/fsm_order.py
max3903: @smangukiya I fixed the logic but we still need to filter the lines and returns with draft ones.

fieldservice_stock/models/fsm_equipment.py
```diff
@@ -0,0 +1,23 @@
...
+
+from odoo import api, fields, models
+
+
+class FSMEquipment(models.Model):
+    _inherit = 'fsm.equipment'
+    _description = "FSM Equipment Inherits"
+
+    lot_id = fields.Many2one(
+        'stock.production.lot', string='Lot/Serial #')
```
max3903: @smangukiya No need for "Lot/". We need a `product_id` field to filter the lot_id. Both fields should be required.

smangukiya:
@max3903 : Source has been improved as per your comments:
1. No need for "Lot/". We need a product_id field to filter the lot_id. Both fields should be required
2. I fixed the logic but we still need to filter the lines and returns with draft ones.
3. current_location_location_id should be set to the current location of the serial number: when the serial number is delivered to the customer location, the current location should change to Partner Location / Customers.

max3903 (approved):

Issue #82, opened 2018-12-31
geospatial for odoo 11


jesulo:
Hi, 

For odoo 11 there is no base_geoengine module, right? because I have not found in

Please refer to the installation instructions available at: https://github.com/OCA/geospatial/tree/11.0/base_geoengine

Regards,

tate11:
https://github.com/OCA/geospatial/pull/178

jesulo:
Thanks @tate11,
yes is in https://github.com/OCA/geospatial/tree/17b429442bc53b03473d347f91f593a0f9be41ad

I have an error, of record

Error:
Odoo Server Error

Traceback (most recent call last):
  File "/home/user99/Developments/odoo11/odoo/odoo/tools/convert.py", line 744, in parse
    self._tags[rec.tag](rec, de, mode=mode)
  File "/home/user99/Developments/odoo11/odoo/odoo/tools/convert.py", line 654, in _tag_record
    id = self.env(context=rec_context)['ir.model.data']._update(rec_model, self.module, res, rec_id or False, not self.isnoupdate(data_node), noupdate=self.isnoupdate(data_node), mode=self.mode)
  File "/home/user99/Developments/odoo11/odoo/odoo/addons/base/ir/ir_model.py", line 1476, in _update
    record = record.create(values)
  File "/home/user99/Developments/odoo11/odoo/odoo/models.py", line 3372, in create
    record = self.browse(self._create(old_vals))
  File "/home/user99/Developments/odoo11/odoo/odoo/models.py", line 3450, in _create
    self._check_selection_field_value(name, val)
  File "/home/user99/Developments/odoo11/odoo/odoo/models.py", line 2104, in _check_selection_field_value
    field.convert_to_cache(value, self)
  File "/home/user99/Developments/odoo11/odoo/odoo/fields.py", line 1808, in convert_to_cache
    raise ValueError("Wrong value for %s: %r" % (self, value))
ValueError: Wrong value for geoengine.vector.layer.classification: 'custom'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user99/Developments/odoo11/odoo/odoo/http.py", line 650, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/user99/Developments/odoo11/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
[...]

PR #59, opened 2018-12-14
[IMP]12573: make custom color for fsm stage will reflect to geoengine mapping

smangukiya:
Modified module:
> fieldservice

Included features:
1. Custom color support for geoengine mapping
2. Increase marker/pointer size in base_geoengine mapping
3. Custom color support for gantt/web_timeline view with fsm.stage color

Note: This change will only work once we include PR https://github.com/asaunier/geospatial/pull/1 with base_geoengine repo.

PR #85, opened 2019-01-02
[FIX] order geo view

brian10048:
@max3903, this is in reference to an error @jesulo reported in #82

### DIFF
--- a/fieldservice/views/fsm_order.xml
+++ b/fieldservice/views/fsm_order.xml
@@ -373,7 +373,7 @@
     <record id="geoengine_vector_layer_fsmordertatecoloredcustom0" model="geoengine.vector.layer">
       <field name="geo_field_id" ref="fieldservice.field_fsm_order_shape"/>
       <field name="name">FSM Order State colored custom</field>
-      <field name="classification">custom</field>
+      <field name="classification">unique</field>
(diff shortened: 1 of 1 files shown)


bodedra (approved):
Reasonable :+1:

jesulo:
Thanks! @brian10048

smangukiya:
@max3903/ @brian10048  : If you take a look at PR https://github.com/OCA/field-service/pull/59. I have mention it  it would work accurately once we merge related https://github.com/asaunier/geospatial/pull/1 

By doing that, this error/issue will no longer be there. As it was developed using that base_geoengine module

I hope this will helps. Thanks

brian10048:
@smangukiya OK, I see what's going on now.  It looks like what we really need is the 11.0 base_geoengine module to be finished and merged into the OCA so we are all using the same one.  In #86 @jesulo is having another issue which I am assuming is related to using a different base_geoengine module.

wolfhall:
@brian10048 - I reached out to CampToCamp to push the GeoEngine V11.0 base_geoengine release out.  Hopefully, they can do this soon. 
@yvaucher

max3903:
https://github.com/OCA/geospatial/pull/178

Issue #89, opened 2019-01-03
fieldservice


max3903:
# res.partner
- [x] Add a many2one field "Primary Service Location" (service_location_id) to fsm.location

# fsm.location
- [x] Add a smart button "Contacts" to show the contacts related to and children of this location
- [x] Add a smart button "Equipments" to show the equipments at this location and sublocations. The smart button only appears to the "Manage Equipments" group.

PR #102, opened 2019-01-15
[IMP] fieldservice

osi-scampbell:
Implemented smart buttons to keep track of partner and equipment count of a location and its sublocations

#89


fieldservice/views/fsm_location.xml
```diff
@@ -22,7 +22,22 @@
         <field name="arch" type="xml">
             <form string="Location">
                 <sheet>
-                    <div class="oe_button_box" name="button_box"/>
+                    <div class="oe_button_box" position="inside">
+                        <button type="object"
+                                name="action_view_equipment"
+                                class="oe_stat_button"
+                                icon="fa-briefcase"
+                                attrs="{'invisible': [('equipment_count', '=', 0)]}" groups="base.group_user">
```
max3903: Should be visible even if count is 0.

Groups should be set to the equipment group.

fieldservice/views/fsm_location.xml
```diff
@@ -22,7 +22,22 @@
...
+                            <field name="equipment_count" widget="statinfo" string="Equipment"/>
+                        </button>                        
+                        <button type="object"
+                                name="action_view_contacts"
+                                class="oe_stat_button"
+                                icon="fa-user-circle-o"
+                                attrs="{'invisible': [('contact_count', '=', 0)]}" groups="base.group_user">
```
max3903: Should be visible even if count is 0.

max3903 (changes requested):
Minor UI changes

fieldservice/views/fsm_location.xml
osi-scampbell: In fieldservice_stock/views/fsm_order.xml line 22 and line 15 do not display their smart buttons when the count is 0. I will change mine, I just wanted to make sure you were aware in case that is an inconsistency.

wolfhall (approved):

PR #118, opened 2019-01-25
Scampbell notes

osi-scampbell:
Copy notes from FSM Location and FSM Equipment when they are selected on a Service Order

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -211,6 +211,21 @@ class FSMOrder(geo_model.GeoModel):
                 timedelta(hours=self.scheduled_duration)
             self.scheduled_date_end = str(date_to_with_delta)
 
+    def copy_notes(self):
+        self.description = ""
+        if self.equipment_id:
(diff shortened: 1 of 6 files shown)


fieldservice/models/fsm_order.py
```diff
@@ -211,20 +211,42 @@ def onchange_scheduled_duration(self):
...
+    def onchange_equipment_id(self):
+        self.copy_notes()
 
     @api.onchange('template_id')
     def _onchange_template_id(self):
         if self.template_id:
             self.category_ids = self.template_id.category_ids
             self.scheduled_duration = self.template_id.hours
-            self.description = self.template_id.instructions
+            self.description += (self.template_id.instructions + '\n')
```
max3903: @osi-scampbell Instructions from the template should be copied in the `todo` field of the order.

max3903: ...and the description should be a text field not a html field.

wolfhall (approved):

PR #123, opened 2019-01-28
[FIX]fieldservice fsm.order view error fix

smangukiya:


smangukiya:
@max3903 : Can you please merge this quick view fix? Thanks

PR #163, opened 2019-03-27
 [MIG][WIP][12.0] fieldservice

murtuzasaleh:
WIP


murtuzasaleh:
@max3903 When I Created this branch and apply the format-patch it gives me the conflict in v12 so I applied format-patch and replaced fieldservice v11 module in my fieldservice_v12 branch and it gives me many [diff.](https://github.com/OCA/field-service/pull/163/commits/8fdcc5c99aef237adbfa35e7a9c81e72363b6a67)

Is it the right way?

max3903:
Yes, this looks good. Can you fix the Travis errors so @osi-scampbell can review and test?

You need to add the oca_dependencies.txt file with web and geospatial in it.

murtuzasaleh:
> Yes, this looks good. Can you fix the Travis errors so @osi-scampbell can review and test?
> 
> You need to add the oca_dependencies.txt file with web and geospatial in it.

Migration is WIP

murtuzasaleh:
@yvaucher I faced an issue in (fsm.location and fsm.order) models when a fieldservice module is installed or update both the table is not created in a database [class FSMLocation(geo_model.GeoModel):
](https://github.com/OCA/field-service/blob/4df3dddc9f6e33338beecd985e9f3eb8e07f62c5/fieldservice/models/fsm_location.py#L12)

I have traced [base_geoengine/geo_model.py](https://github.com/OCA/geospatial/pull/185/files#diff-758bcd153189712715dfd56ddc0c293dL23) this file and found that when I pass _register = False the next error is Model 'fsm.location' does not exist in registry in V12

Is this because of a base_geoengine module? Please correct me if I am wrong.

@max3903

max3903:
@murtuzasaleh Please comment the dependency on base_geoengine for now and move forward...

osi-scampbell:
The module installs and runs cleanly on my end so far. To fix the travis we need to add an oca_dependency.txt file like this one here in v11 https://github.com/OCA/field-service/blob/11.0/oca_dependencies.txt. We also need to add the postgis extensions to the database before installing which we can do through the travis.yml file, see https://github.com/OCA/field-service/blob/11.0/.travis.yml lines 19, 20 and 21

murtuzasaleh:
> The module installs and runs cleanly on my end so far. To fix the travis we need to add an oca_dependency.txt file like this one here in v11 https://github.com/OCA/field-service/blob/11.0/oca_dependencies.txt. We also need to add the postgis extensions to the database before installing which we can do through the travis.yml file, see https://github.com/OCA/field-service/blob/11.0/.travis.yml lines 19, 20 and 21

Sure, I will add it after module migration part is done.

fieldservice/__manifest__.py
```diff
@@ -11,6 +11,7 @@
     'depends': [
         'base_geolocalize',
         'mail',
+        'stock',
```
max3903: Why are you adding this dependency? We don't want to depend on stock in this module.

murtuzasaleh:
@max3903 As per our discussion I have removed base_geoengine related code in this branch, can you please review it.

I will create fieldservice_geoengine PR soon.

fieldservice/wizard/fsm_wizard.py
```diff
@@ -0,0 +1,42 @@
...
+            self._context.get('active_ids', []))
+        for partner in partners:
+            if self.fsm_record_type == 'person':
+                self.action_convert_person(partner)
+            if self.fsm_record_type == 'location':
+                self.action_convert_location(partner)
+        return {'type': 'ir.actions.act_window_close'}
+
+    def action_convert_location(self, partner):
+        pass
```
max3903: @murtuzasaleh Even if stock is not installed, we still need to convert the partner record into a fsm.location.

oca_dependencies.txt
```diff
@@ -0,0 +1,2 @@
+web
+
```
max3903: Empty line.

fieldservice/wizard/fsm_wizard.py
```diff
@@ -29,7 +29,24 @@ def action_convert(self):
...
+                        'customer_id': partner.id}
+                self.env['fsm.location'].create(vals)
+                partner.write({'fsm_location': True})
+            else:
+                vals = {'partner_id': partner.id,
+                        'owner_id': partner.id,
+                        'customer_id': partner.id}
+                self.env['fsm.location'].create(vals)
+                partner.write({'fsm_location': True})
+        else:
```
max3903, line 40: Same code in the if and the else so:
```
        if res == 0:
            vals = {'partner_id': partner.id,
                    'owner_id': partner.id,
                    'customer_id': partner.id}
            self.env['fsm.location'].create(vals)
            partner.write({'fsm_location': True})
        else:
```

murtuzasaleh: Done

max3903:
@murtuzasaleh Please squash your commits.

Issue #162, opened 2019-03-19
[IMP] fieldservice: multi-company


max3903:
# Todo

- [x] Add a many2one field "Company" (company_id) to:
  - [x] fsm.order
  - [x] fsm.vehicle
  - [x] fsm.equipment
  - [x] fsm.stage
  - [x] fsm.team
  - [x] territory
  - [x] branch
  - [x] district
  - [x] region

- [x] Add the field to the form views and display it only to the multi-company group
- [x] The field defaults to the company of the logged in user.
- [x] Add global record rules for the following objects to allow users to only read the records of their company and its children:

  - [x] fsm.order: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] fsm.vehicle: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] fsm.equipment: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] fsm.location: `['|',('partner_id.company_id','=',False),('partner_id.company_id','child_of',[user.company_id.id])]`
  - [x] fsm.person: `['|',('partner_id.company_id','=',False),('partner_id.company_id','child_of',[user.company_id.id])]`
  - [x] fsm.stage: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] fsm.team: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] territory: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] branch: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] district: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`
  - [x] region: `['|',('company_id','=',False),('company_id','child_of',[user.company_id.id])]`

PR #165, opened 2019-04-01
[IMP] multi-company support

brian10048:
@max3903 Please review for #162 

I followed your spec.  I'm wondering if we should we also setup multi-company support on the FSM category and FSM order template models?


osi-scampbell (approved):
Everything looks good! If I had to make one suggestion it would be on fsm_territory. Here we have the company_id field in the right column but on fsm_region, fsm_district, and fsm_branch the company_id field is in the left column. For consistency purposes should we try to keep that field in the same column of each form view?

brian10048:
I agree. I modified to keep consistency with other modules of Odoo and moved company fields to right hand column in views

max3903 (approved):

murtuzasaleh (approved):

fieldservice/security/company_security.xml
```diff
@@ -0,0 +1,73 @@
+<odoo noupdate="1">
```
murtuzasaleh, line 2: Please add this line <?xml version="1.0" encoding="utf-8"? >

fieldservice_vehicle/security/vehicle_security.xml
```diff
@@ -0,0 +1,10 @@
+<odoo noupdate="1">
```
murtuzasaleh, line 2: Please add this line <?xml version="1.0" encoding="utf-8"? >

murtuzasaleh:
LGTM @max3903 Can you please merge this PR.

PR #174, opened 2019-04-04
[MIG][12.0] fieldservice_stock

murtuzasaleh:
Depends on:
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/556


murtuzasaleh:
@max3903 Added this code in this [prepare_equipment_values](https://github.com/OCA/field-service/blob/557cf0f8e881b1e5fca4a2cad1f34063135d708a/fieldservice_stock/models/stock.py#L16) method, because found error of current_location_id while validate stock flow.

Is it the right way or I will remove required equal to True attribute from [current_location_id](https://github.com/OCA/field-service/blob/f97e337ea53a13de3c3396ee977ea613cdc9d057/fieldservice/models/fsm_equipment.py#L20) fields in .py file and set it in .xml file?

fieldservice_stock/models/stock.py
```diff
@@ -0,0 +1,129 @@
...
+                                     move_line.lot_id.name),
+                'product_id': move_line.product_id.id,
+                'lot_id': move_line.lot_id.id,
+                'current_location_id': move_line.picking_id and
+                move_line.picking_id.partner_id and
+                move_line.picking_id.partner_id and
+                move_line.picking_id.partner_id.
+                service_location_id and
+                move_line.picking_id.partner_id.
+                service_location_id.id or False}
```
max3903: @murtuzasaleh The equipment should be created with the destination location. No need to check the primary location of the partner:

```
def prepare_equipment_values(self, move_line):
    return {'name': '%s (%s)' % (move_line.product_id.name, move_line.lot_id.name),
            'product_id': move_line.product_id.id,
￼            'lot_id': move_line.lot_id.id,
￼            'current_location_id': move_line.request_id.fsm_order_id.location_id}
￼            'current_stock_location_id': move_line.dest_location_id}
```

fieldservice_stock/models/stock.py
```diff
@@ -0,0 +1,129 @@
...
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    fsm_order_line_id = fields.Many2one('fsm.order.line', 'FSM Order Line')
+    fsm_order_return_line_id = fields.Many2one(
+        'fsm.order.return', 'FSM Order Return Line')
```
max3903: `fsm.order.line` and `fsm.order.return` should be replaced by stock requests.

fieldservice_stock/models/stock.py
murtuzasaleh: @max3903 This PR is WIP because as per mentioned in the wiki (Migrate to use stock_request instead of fsm.order.line and fsm.order.return) So I am working on it.

murtuzasaleh:
@max3903 stock_request related PR is still WIP, I am working on it, So can you please review [12.0][MIG] PR.
I have mentioned in all the PR which are ready to review.

murtuzasaleh:
@max3903 In V11 Inbound functionality is working?
In my side, Inbound functionality is not working.

max3903:
@murtuzasaleh It's the same feature as outbound but in the other direction (Customer Location --> Stock). We want to use `stock.request` as well.

murtuzasaleh:
> @murtuzasaleh It's the same feature as outbound but in the other direction (Customer Location --> Stock). We want to use `stock.request` as well.

@max3903 So can we create only one one2many (stock.request) field and in the one2many line will add one type field (outbound and Inbound) based on that route_id will appear (In Route 2 boolean field In v11 FSM order line and FSM return line) so based on the onchange domain is a change in the route_id field. 

Is it fine?  If not so I want to create two models for (outbound and Inbound) and same for button code.

max3903:
@murtuzasaleh Yes, let's go with one field, but be careful the route is hidden to the FSM user and only visible by the Inventory one.

Instead of naming the field `type`, let's name it `direction` with 'outbound' and 'inbound'.

murtuzasaleh:
> @murtuzasaleh Yes, let's go with one field, but be careful the route is hidden to the FSM user and only visible by the Inventory one.
> 
> Instead of naming the field `type`, let's name it `direction` with 'outbound' and 'inbound'.

Sure, Thanks

murtuzasaleh:
@max3903 @osi-scampbell Can you please review this PR.

fieldservice_stock/models/stock.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (C) 2018 - TODAY, Brian McMaster
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class StockRequest(models.Model):
+    _inherit = "stock.request"
+
+    return_order_id = fields.Many2one(
```
max3903:
```suggestion
    fsm_order_id = fields.Many2one(
```

fieldservice_stock/models/stock_production_lot.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockProductionLot(models.Model):
+    _inherit = 'stock.production.lot'
+
+    equipment_id = fields.Many2one('fsm.equipment',
```
max3903:
```suggestion
    fsm_equipment_id = fields.Many2one('fsm.equipment',
```

murtuzasaleh: @max3903 Done

max3903 (approved):

PR #201, opened 2019-04-29
[ADD] fieldservice_purchase

osi-scampbell:
Odoo Purchases App and the Field Service App do not currently have a way to link FSM Workers
to their corresponding Pricelist.

This module allows you to link a Field Service Worker back to their corresponding Vendor Pricelist(s).


fieldservice_purchase/models/fsm_person.py
```diff
@@ -0,0 +1,35 @@
...
+    pricelist_count = fields.Integer(
+        compute='_compute_pricelist_count',
+        string='# Pricelists'
+    )
+
+    @api.multi
+    def _compute_pricelist_count(self):
+        for worker in self:
+            worker.pricelist_count = self.env['product.supplierinfo'].\
+                search_count([('name', '=', worker.name)])
```
max3903:
```suggestion
                search_count([('id', '=', worker.partner_id)])
```

fieldservice_purchase/models/fsm_person.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @api.multi
+    def action_view_pricelists(self):
+        for worker in self:
+            pricelist = self.env['product.supplierinfo'].search(
+                [('name', '=', worker.name)])
```
max3903:
```suggestion
                [('id', '=', worker.partner_id)])
```

fieldservice_purchase/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'license': 'AGPL-3',
+    'version': '12.0.1.0.0',
+    'depends': [
+        'fieldservice',
+        'purchase',
+    ],
+    'data': [
+        'views/fsm_person.xml',
+    ],
+    'installable': True,
```
max3903:
```suggestion
    'development_status': 'Beta',
```

For the installation, please add an option in the settings panel.

murtuzasaleh: @max3903 `For the installation, please add an option in the settings panel.` where?

max3903: In fieldservice:
https://github.com/OCA/field-service/blob/12.0/fieldservice/views/res_config_settings.xml

fieldservice_purchase/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
max3903, line 2:
```suggestion
# Copyright (C) 2019 - TODAY, Open Source Integrators
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```

max3903 (approved):

fieldservice_purchase/views/fsm_person.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
```
patrickrwilson: This is no long er needed in v12 and the xml is missing copyright/license info.

patrickrwilson (approved):
Small change on xml.

brian10048:
@osi-scampbell  
Can you also make this module enable the `group_manage_vendor_price` from Purchase settings in order to enable the pricelist menus in Purchases app by default?

On one of the runbot builds I created a new FSM Worker from Field Service app > Master Data > Workers menu.  Upon save, the pricelist count goes to 1.  It seems to set a random pricelist of a random vendor? 

Also, do you think this should the `supplier` field to True of the `res.partner` model associated to the FSM worker?

patrickrwilson:
@brian10048 
> Also, do you think this should the `supplier` field to True of the `res.partner` model associated to the FSM worker?

I don't think this is a good idea because not every worker will be a supplier/vendor, some would be employees or ones that may not be classified as a supplier or vendor. However its probably a good idea to place the Is Vendor book field on the worker form to easily check that option if it's not already done so in the code.

brian10048:
> 
> > Also, do you think this should the `supplier` field to True of the `res.partner` model associated to the FSM worker?
> 
> I don't think this is a good idea because not every worker will be a supplier/vendor, some would be employees or ones that may not be classified as a supplier or vendor. However its probably a good idea to place the Is Vendor book field on the worker form to easily check that option if it's not already done so in the code.

@patrickrwilson You are right, a field on the FSM worker would be a better option

PR #212, opened 2019-05-20
[IMP][12.0] fieldservice_delivery: Improved onchange method.

murtuzasaleh:
Improved onchange method.


murtuzasaleh:
@max3903 @osi-scampbell Can you please review this PR.

max3903 (changes requested):
@murtuzasaleh Choosing Fedex services is customer-specific.

PR #214, opened 2019-05-20
[IMP] fieldservice: Add an option to auto-fill workers on location

max3903:


patrickrwilson (approved):
Tested and works as intended. Since field service is still a work in progress, i'm not sure the process of verifying travis errors are resolved. Might want to resolve those if it's required to be fully green before merging.

brian10048 (approved):
Have not had a chance to test.  I like the idea and code looks good

PR #215, opened 2019-05-20
[IMP] fieldservice: Sort workers on location

max3903:
- [x] Depends on #214

### DIFF
--- a/fieldservice/__manifest__.py
+++ b/fieldservice/__manifest__.py
@@ -33,6 +33,7 @@
         'views/fsm_tag.xml',
         'views/res_partner.xml',
         'views/fsm_location.xml',
+        'views/fsm_location_person.xml',
         'views/fsm_person.xml',
         'views/fsm_order.xml',
         'views/fsm_route.xml',
--- a/fieldservice/models/__init__.py
+++ b/fieldservice/models/__init__.py
@@ -15,6 +15,7 @@ from . import (
     fsm_team,
     fsm_person,
     fsm_location,
+    fsm_location_person,
     res_partner,
     res_state,
     res_country,
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -44,11 +44,9 @@ class FSMLocation(models.Model):
     fsm_parent_id = fields.Many2one('fsm.location', string='Parent',
                                     index=True)
     notes = fields.Text(string="Notes")
-    person_ids = fields.Many2many('fsm.person',
-                                  'fsm_person_location_rel',
-                                  'fsm_location_id',
-                                  'fsm_person_id',
-                                  string='Preferred Workers')
+    person_ids = fields.One2many('fsm.location.person',
+                                 'location_id',
+                                 string='Workers')
     contact_count = fields.Integer(string='Contacts Count',
                                    compute='_compute_contact_ids')
     equipment_count = fields.Integer(string='Equipment',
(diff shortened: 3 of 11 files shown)


patrickrwilson:
@max3903 if you add several workers onto a location in the preferred workers page, then you create an order, then try to add a requested order, those workers don't show in the list. I'm not sure if the filtering works properly on that list so that would be something to check too.
image

image

max3903:
@patrickrwilson Fixed.

osi-scampbell (approved):

patrickrwilson:
> @patrickrwilson Fixed.

@max3903 it looks like the filter was removed entirely. Ideally, when you go to add a preferred worker on the order, it has a default filter to show only the preferred workers related to the location on the order. That filter should be removable to search all other workers. 

image

patrickrwilson:
@max3903 Another suggestion, since fsm.location.person is a new model, perhaps a menu item in 'Master Data->Worker Locations' is useful to be able to import and manage the larger list. This menu item can be placed directly under 'Worker Skills'

max3903:
Ping @smangukiya

max3903:
> @max3903 Another suggestion, since fsm.location.person is a new model, perhaps a menu item in 'Master Data->Worker Locations' is useful to be able to import and manage the larger list. This menu item can be placed directly under 'Worker Skills'

Included here: https://github.com/OCA/field-service/pull/215/files#diff-1b58a0fbe64ef98c07dc251c1f902189R86

patrickrwilson:
> > @max3903 Another suggestion, since fsm.location.person is a new model, perhaps a menu item in 'Master Data->Worker Locations' is useful to be able to import and manage the larger list. This menu item can be placed directly under 'Worker Skills'
> 
> Included here: https://github.com/OCA/field-service/pull/215/files#diff-1b58a0fbe64ef98c07dc251c1f902189R86

I think that setting is for auto populating the workers by the territory that is selected on the location during creation. The menu item would show all the fsm.location.person, which would be very similar to the worker skills, but shows the location -> worker records.

patrickrwilson:
> > @patrickrwilson Fixed.
> 
> @max3903 it looks like the filter was removed entirely. Ideally, when you go to add a preferred worker on the order, it has a default filter to show only the preferred workers related to the location on the order. That filter should be removable to search all other workers.
> 
> image

@max3903 looks like it's closer after your last commit, but doesn't show all the workers on the location. I setup a sample location and added three workers, but only one shows when trying to add preferred workers to an order.

max3903:
@patrickrwilson Sorry, it is not meant to work yet.

The context on the fsm.order view provides the ID of the location and use it to search in the id column of fsm.location.person, not in the location_id column... 

...and `{'search_default_locations_ids.location_id': location_id}` does not work :(

@smangukiya will look into fixing this on friday.

smangukiya:
@max3903 : I made an improvement at https://github.com/OCA/field-service/pull/223. 
Thanks
S

patrickrwilson:
@max3903 i just tried it again, added two preferred workers but when i try to add them to an order, the filtering isn't showing them.
image

smangukiya:
> @max3903 i just tried it again, added two preferred workers but when i try to add them to an order, the filtering isn't showing them.
> image

@patrickrwilson : It seems like fix is not merge yet to 12.0 branch yet. Once, it will be merge you will be able to see default filtered list of preferred workers for that location.

CC: @max3903

patrickrwilson (approved):

Issue #187, opened 2019-04-18
V 11 Day routes - Expected Singleton


sapentia:
Hi

testing the v11 verison on a clean install ubuntu 16.04. 
create 2 orders. 
create a day route record
add first to the day route, ok
attempt to add second one to Day Routes record. I get a 'expected singleton' error. 
is this an issue or just not completed yet?

brian10048:
This is an issue.  I have duplicated the error by following similar steps.

Here is the traceback received when attempting to Save the Day Route. 

Error:
Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/odoo11/odoo/http.py", line 651, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo11/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/opt/odoo/odoo11/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/opt/odoo/odoo11/odoo/http.py", line 693, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/odoo11/odoo/http.py", line 342, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/odoo11/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/odoo11/odoo/http.py", line 335, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/odoo11/odoo/http.py", line 937, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/odoo11/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/odoo11/addons/web/controllers/main.py", line 934, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/opt/odoo/odoo11/addons/web/controllers/main.py", line 926, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo11/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/opt/odoo/odoo11/odoo/api.py", line 672, in call_kw_model
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/field-service/fieldservice/models/fsm_route.py", line 32, in create
[...]

PR #222, opened 2019-05-26
[12.0][FIX] Issue 187: Expected Singleton

bodedra:
Issue Reference : #187

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -199,9 +199,11 @@ class FSMOrder(models.Model):
                 vals.get('scheduled_date_start')) + \
                 timedelta(hours=self.scheduled_duration)
             vals['scheduled_date_end'] = str(date_to_with_delta)
-        if 'customer_id' not in vals:
-            vals['customer_id'] = self.location_id.customer_id.id
-        return super(FSMOrder, self).write(vals)
+        res = super(FSMOrder, self).write(vals)
+        for order in self:
+            if 'customer_id' not in vals and order.customer_id is False:
+                order.customer_id = order.location_id.customer_id.id
+        return res
 
     def action_confirm(self):
         return self.write({'stage_id': self.env.ref(


bodedra:
Would you please review these.

@max3903 @brian10048 @kittiu

kittiu:
@bodedra I have test it. There is no error now. 

image

But with the above code, all fsm.order in self would be written with the same customer_id (from last order), when they might not be?

Or should it be something like,

image

bodedra:
@kittiu May you test it again ? I have improve logic.

If these approach will fail, will try with yours.

kittiu:
I think it loop through orders but still write in self multiple time? I am not sure that would work.

bodedra:
@kittiu Odoo hr_recruitment module has issued same technique . Here is reference https://github.com/odoo/odoo/blob/12.0/addons/hr_recruitment/models/hr_recruitment.py#L284

kittiu:
@bodedra yes, I have seen it. Is it possible that the hr_recruitment is also not correct? But when it try to update **last_stage_id** may be it is ok as to use the stage from the last applicant?
Same here, I am not sure it is ok to write using value from the last fsm.order? (I really don't know what is expected in term functional)

Note: I think within the for loop it will be using the value from cache to write to the records. If that so, it will be equal with write to all orders with the value from the last order?

May be we need other to have second opinion?

bodedra:
@kittiu That's my bad. Please ignore hr_recruitment. :smile:  Yes. It updates last location customer in all FSM order. I have tried locally with your suggested code and work fine. :+1:

max3903:
@bodedra 

```
@api.multi
def write(self, vals)
    for order in self:
        ...
    return super(FSMOrder, self).write(vals)
```

bodedra:
> @bodedra
> 
> ```
> @api.multi
> def write(self, vals)
>     for order in self:
>         ...
>     return super(FSMOrder, self).write(vals)
> ```

@max3903 These will not take updated FSM location. So IMO our current fix will be best for updated location with their customer.

fieldservice/models/fsm_order.py
```diff
@@ -199,9 +199,11 @@ def write(self, vals):
                 vals.get('scheduled_date_start')) + \
                 timedelta(hours=self.scheduled_duration)
             vals['scheduled_date_end'] = str(date_to_with_delta)
+        res = super(FSMOrder, self).write(vals)
         if 'customer_id' not in vals:
-            vals['customer_id'] = self.location_id.customer_id.id
-        return super(FSMOrder, self).write(vals)
+            for fsm_order in self:
+                fsm_order.customer_id = fsm_order.location_id.customer_id.id
+        return res
```
max3903, line 206:
```
        res = super(FSMOrder, self).write(vals)
        for order in self:
            if 'customer_id' not in vals and order.customer_id is False:
                order.customer_id = order.location_id.customer_id.id
        return res
```

bodedra:
@max3903 Thanks for your suggestion. Comment attended!

max3903 (approved):

brian10048 (approved):

PR #223, opened 2019-05-29
[IMP]16335: fieldservice improvement to support preferred worker filer based on selected location #215

smangukiya:
Depends on #215

### DIFF
--- a/fieldservice/models/fsm_person.py
+++ b/fieldservice/models/fsm_person.py
@@ -25,6 +25,27 @@ class FSMPerson(models.Model):
     territory_ids = fields.Many2many('fsm.territory', string='Territories')
 
     @api.model
+    def _search(self, args, offset=0, limit=None, order=None, count=False,
+                access_rights_uid=None):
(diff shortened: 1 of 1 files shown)


max3903:
@smangukiya Can we avoid the SQL query and use the ORM? Can you fix the typo: `preferred_worker_list`?

max3903:
@smangukiya Can you make this pull request (or a new one) against this branch:
https://github.com/ursais/field-service/tree/12.0-fix-pm-516

So we can test everything together before merging here: #215

smangukiya:
> @smangukiya Can we avoid the SQL query and use the ORM? Can you fix the typo: `preferred_worker_list`?

@max3903 : I have corrected typo, but with regards to many2many field and it would be much faster to process preferred worker list I would suggest to go with sql query directly.

Thanks

osi-scampbell (approved):

patrickrwilson (approved):

max3903 (approved):

PR #224, opened 2019-05-29
[WIP][ADD] Field Service Stage Server Action

osi-scampbell:
This PR adds Field Service Stage Server Action to the repo. This module adds action_id to a fsm_stage record so the user can specify what emails should be sent from this stage. There are also helper functions on fsm_order so the ir_server_actions know when stage_id is changed.


max3903:
Superseeded by #234

PR #232, opened 2019-06-06
[FIX] fieldservice_agreement

osi-scampbell:
This PR adds notes to Service Profiles, adds a tab on Agreements where users can add Service Profiles, adds tab on FSM Location where users can see the Agreements associated with this location and their Service Profiles. Tickets #: 609, 608, 604

### DIFF
--- a/fieldservice_agreement/models/fsm_location.py
+++ b/fieldservice_agreement/models/fsm_location.py
@@ -1,0 +1,26 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class FSMLocation(models.Model):
+    _inherit = 'fsm.location'
+
+    serviceprofile_ids = fields.Many2many('agreement.serviceprofile',
+                                          string="Service Profiles",
+                                          compute='_compute_service_ids')
+
+    @api.multi
+    def _compute_service_ids(self):
+        for loc in self:
+            agreements = self.env['agreement'].\
+                search([('fsm_location_id', '=', self.name)])
+            ids = []
+            for agree in agreements:
+                servpros = self.env['agreement.serviceprofile'].\
+                    search([('agreement_id', '=', agree.id)])
+                for ser in servpros:
+                    if ser.id not in ids:
+                        ids.append(ser.id)
+            self.serviceprofile_ids = ids
--- a/fieldservice_agreement/views/fsm_location.xml
+++ b/fieldservice_agreement/views/fsm_location.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="fsm_location_agreement_serviceprofile" model="ir.ui.view">
+        <field name="name">fsm.location.agreement.serviceprofile</field>
+        <field name="model">fsm.location</field>
+        <field name="inherit_id" ref="fieldservice.fsm_location_form_view"/>
+        <field name="arch" type="xml">
+            <notebook position="inside">
+                <page string="Service Profiles">
+                    <field name="serviceprofile_ids">
+                        <tree>
+                            <field name="name"/>
+                            <field name="agreement_id"/>
+                        </tree>
+                    </field>
+                </page>
+            </notebook>
+        </field>
+    </record>
+</odoo>
--- a/fieldservice_agreement/views/fsm_location.xml
+++ b/fieldservice_agreement/views/fsm_location.xml
@@ -1,4 +1,5 @@
-<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 Open Source Integrators
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl) -->
 <odoo>
 
     <record id="fsm_location_agreement_serviceprofile" model="ir.ui.view">
@@ -18,4 +19,5 @@
             </notebook>
         </field>
     </record>
+
 </odoo>
--- a/fieldservice_agreement/views/fsm_location.xml
+++ b/fieldservice_agreement/views/fsm_location.xml
@@ -13,6 +13,7 @@
                         <tree>
                             <field name="name"/>
                             <field name="agreement_id"/>
+                            <field name="notes"/>
                         </tree>
                     </field>
                 </page>
--- a/fieldservice_agreement/__manifest__.py
+++ b/fieldservice_agreement/__manifest__.py
@@ -18,7 +18,8 @@
         'views/fsm_order_view.xml',
         'views/fsm_equipment_view.xml',
         'views/agreement_view.xml',
-        'views/fsm_person.xml'
+        'views/fsm_person.xml',
+        'views/fsm_location.xml'
     ],
     'installable': True,
     'development_status': 'Beta',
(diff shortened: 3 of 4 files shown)


fieldservice_agreement/models/agreement.py
```diff
@@ -58,3 +62,13 @@ def action_view_fsm_equipment(self):
             else:
                 action['domain'] = [('id', 'in', equipment_ids.ids)]
             return action
+
+    @api.multi
+    def _set_service_ids(self):
+        for agree in self:
+            servpros = self.env['agreement.serviceprofile'].\
+                search([('agreement_id', '=', self.name)])
```
max3903:
```suggestion
                search([('agreement_id', '=', self.id)])
```

fieldservice_agreement/views/agreement_serviceprofile.xml
```diff
@@ -0,0 +1,19 @@
...
+        <field name="type">form</field>
+        <field name="inherit_id" ref="agreement_serviceprofile.agreement_serviceprofile_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="/form/sheet/group/group[2]" position="after">
+                <group string="Notes">
+                    <field name="notes" nolabel="1" widget="html"/>
+                </group>
+            </xpath>
+            <field name="agreement_id" position="attributes">
+                <attribute name="context">{'default_agreement_id': agreement_id}</attribute>
```
max3903: @osi-scampbell This doesn't make sense. You are selecting the agreement field and the context should have other fields to help you select the right agreement.

fieldservice_agreement/models/agreement_serviceprofile.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import fields, models
+
+
+class AgreementServiceProfile(models.Model):
+    _inherit = 'agreement.serviceprofile'
+
+    notes = fields.Text(string="Notes")
```
max3903: @osi-scampbell Should be moved to `agreement_serviceprofile`.

fieldservice_agreement/models/agreement.py
```diff
@@ -15,6 +15,10 @@ class Agreement(models.Model):
                                      compute='_compute_equipment_count')
     fsm_location_id = fields.Many2one('fsm.location', string="FSM Location")
 
+    serviceprofile_ids = fields.Many2many('agreement.serviceprofile',
+                                          default='_set_service_ids',
+                                          string="Service Profiles")
```
max3903: @osi-scampbell Should be moved to `agreement_serviceprofile` if not already there.

fieldservice_agreement/models/agreement.py
```diff
@@ -58,3 +62,13 @@ def action_view_fsm_equipment(self):
...
+
+    @api.multi
+    def _set_service_ids(self):
+        for agree in self:
+            servpros = self.env['agreement.serviceprofile'].\
+                search([('agreement_id', '=', self.name)])
+            ids = []
+            for ser in servpros:
+                ids.append(ser.id)
+            self.serviceprofile_ids = ids
```
max3903: @osi-scampbell You are creating a new agreement. How can you set the service profiles by default based on the service profiles related to that agreement? It's a new one so there is no service profile that are linked to it yet.

fieldservice_agreement/models/fsm_location.py
```diff
@@ -0,0 +1,26 @@
...
+
+    serviceprofile_ids = fields.Many2many('agreement.serviceprofile',
+                                          string="Service Profiles",
+                                          compute='_compute_service_ids')
+
+    @api.multi
+    def _compute_service_ids(self):
+        for loc in self:
+            agreements = self.env['agreement'].\
+                search([('fsm_location_id', '=', self.name)])
```
max3903, line 18:
```suggestion
                search([('fsm_location_id', '=', self.id)])
```

fieldservice_agreement/models/fsm_location.py
```diff
@@ -0,0 +1,26 @@
...
+            ids = []
+            for agree in agreements:
```
max3903, line 20:
```suggestion
            for agreement in agreements:
```

fieldservice_agreement/models/fsm_location.py
```diff
@@ -0,0 +1,26 @@
...
+                servpros = self.env['agreement.serviceprofile'].\
+                    search([('agreement_id', '=', agree.name)])
```
max3903:
```suggestion
                    search([('agreement_id', '=', agreement.id)])
```

fieldservice_agreement/views/agreement_view.xml
```diff
@@ -32,6 +32,14 @@
...
             </div>
             <notebook position="inside">
+                <page name="serviceprofiles" string="Service Profiles">
+                    <field name="serviceprofile_ids" 
+                           context="{'default_agreement_id': active_id}">
+                        <tree>
+                            <field name="name"/>
+                        </tree>
+                    </field>
+                </page>
```
max3903: @osi-scampbell Should be moved to `agreement_serviceprofile` if not already there.

fieldservice_agreement/views/agreement_serviceprofile.xml
osi-scampbell: When I remove the context from this field and create a serviceprofile from that many2many list on an agreement the agreement_id is not automatically populated. I figured that is the functionality we would want.

fieldservice_agreement/models/agreement_serviceprofile.py
osi-scampbell: That was my first thought but I figured if OCA took it out then they wouldn't want it put back in. I'll move it right away.

fieldservice_agreement/models/agreement.py
osi-scampbell: I think the confusion was I had serviceprofiles as Many2many when they should have been One2many. That method worked fine for Many2many but they should be One2many, and if they are then this function is not necessary.

fieldservice_agreement/views/agreement_serviceprofile.xml
max3903: Which many2many? A service profile belongs to only one agreement:
https://github.com/OCA/contract/blob/12.0/agreement_serviceprofile/models/agreement_serviceprofile.py#L13

osi-scampbell: That is correct I had serviceprofiles as Many2many instead of One2many. I changed all that around. However, when you create a serviceprofile from an agreement, the agreement_id does not autopopulate. That was the intention of that code.

max3903 (approved):

fieldservice_agreement/views/fsm_location.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
```
patrickrwilson: <?xml version="1.0" encoding="utf-8"?> Isn't needed anymore for v12.

patrickrwilson (approved):
Please add the service profile notes to the list on locations so you can also see the notes related to the service profile. This allows the user to quickly see details without opening up each profile. Approving this even if that doesn't make it in this update.

Issue #142, opened 2019-02-27
[ADD] fieldservice_stage_server_action


max3903:
Depends: fieldservice
License: AGPL-3

Odoo Field Service App does not provide a way to communicate with third parties within the process of an order.

Organizations may need to communicate with different contacts or workers to notify them, to get information or a decision from them.

This module provides a server action on each stage of the process so you can create child actions to send different email templates to different groups of people. Those actions are executed when you click on the buttons to move the order within the process.

# Configuration

* Go to Settings > Technical > Email > Templates
* Create or select an email template
* Go to Field Service > Configuration > Stages
* Create a or select a stage
* Add or open the server action
* Create a server action as a child action to send an email using the previous template

# Usage

* Go to Field Service
* Create an order and go through the process
* Actions will be executed and email sent as the order move from one stage to another

# Technical Specifications

- [x] Add a many2one field "Server Action" (action_id) to ir.actions.server on fsm.stage
- [ ] Add one server action for each stage:
  * Name = "FSM Order is " + Name of the stage
  * Model = fsm.order
  *  Action To Do = Execute Several Actions
- [x] Execute the server action when an order enters a stage

PR #234, opened 2019-06-09
[ADD] fieldservice_stage_server_action

osi-scampbell:
This PR adds a many2one field for fsm.stage to ir.server.action. The module also has an ir.server.action for each stage when an email should be sent out.

#142


fieldservice/models/fsm_order.py
```diff
@@ -299,6 +299,41 @@ def _onchange_template_id(self):
...
+        email = ""
+        if self.person_id:
+            for person in self.person_ids:
+                if (person.email and (person != self.person_id)):
+                    email += person.email + ','
+        email = email[:-1]
+        self._compute_workers_not_selected = email
+
+    def _compute_outage_list(self):
+        rtype = self.env['res.partner.relation.type'].\
```
max3903: @osi-scampbell `fieldservice` does not depend on the module providing this model. It will not work.

max3903:
@osi-scampbell Where do you make the relationship between a stage and its action?

fieldservice/models/fsm_order.py
osi-scampbell: I see, I think it should be moved to pavlov_media_fieldservice then

osi-scampbell:
> @osi-scampbell Where do you make the relationship between a stage and its action?

In fsm_stage.py I add action_id which is a link to ir.actions.server. When it is installed the user goes to a stage and sets the action there.

fieldservice/models/fsm_order.py
max3903: Yes please.

osi-scampbell: Done

max3903:
@osi-scampbell We should update the list of stages and set the action. See an example below.

in data/fsm_stage.xml:
```
<record id="fieldservice.fsm_stage_confirmed" model="fsm.stage">
    <field name="action_id" ref="fsm_order_parent_confirmed"/>
</record>
```

osi-scampbell:
> @osi-scampbell We should update the list of stages and set the action. See an example below.
> 
> in data/fsm.stage.xml:
> 
> ```
> <record id="fieldservice.fsm_stage_confirmed" model="fsm.stage">
>     <field name="action_id" ref="fsm_order_parent_confirmed"/>
> </record>
> ```

So that manually sets the action_id so the user doesn't have to configure it themselves?

max3903:
@osi-scampbell Yes!

osi-scampbell:
Perfect, there are a few more spots I think that would be helpful in. I will add them now.

max3903 (approved):

fieldservice_stage_server_action/data/fsm_stage.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
```
patrickrwilson: <?xml version="1.0" encoding="utf-8"?> not needed anymore in v12

fieldservice_stage_server_action/data/ir_server_action.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
```
patrickrwilson: <?xml version="1.0" encoding="utf-8"?> not needed anymore in v12

fieldservice_stage_server_action/views/fsm_stage.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
patrickrwilson: <?xml version="1.0" encoding="utf-8"?> not needed anymore in v12

fieldservice_stage_server_action/data/base_automation.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
```
patrickrwilson: <?xml version="1.0" encoding="utf-8"?> not needed anymore in v12

patrickrwilson (approved):
Small changes

PR #235, opened 2019-06-12
[IMP] fieldservice: Filter Assigned like Request Workers (PM #627)

osi-scampbell:
Add the filter that we made for Request Worker onto the Assigned To field


max3903 (approved):

patrickrwilson (approved):
Tested and works good

PR #240, opened 2019-06-17
[IMP] fieldservice_partner_multi_relation: Integrate FSM in Partner Multi Relation

osi-scampbell:
This PR integrates FSM Locations into the Partner Multi Relations module. This module only runs when one of the three fields in a relation type (left, type, right) are connected to an FSM Location. If none of them are, then the base logic handles the relation type. 

When an FSM Location is selected in one of the three, then domains are set to help make sure the user can only select partners or relationships that are compatible. This module requires one small change in workflow and that is relation types must have unique names. 

The base module you can selected relationship types with the same names but different relationships however the user then must guess that relationships left and right partner which I do not believe is ideal. The relationship types in the data are all prefixed with which categories should be on the left partner and the right partner.

### DIFF
--- a/fieldservice_partner_multi_relation/data/demo.xml
+++ b/fieldservice_partner_multi_relation/data/demo.xml
@@ -22,19 +22,19 @@
 
     <!-- manages  -->
     <record id="rel_type_manages_loc" model="res.partner.relation.type">
-        <field name="name">[L+L] manges</field>
+        <field name="name">[L+L] manages</field>
         <field name="name_inverse">[L+L] is managed by</field>
         <field name="contact_type_left">fsm-location</field>
         <field name="contact_type_right">fsm-location</field>
     </record>
     <record id="rel_type_manages_org" model="res.partner.relation.type">
-        <field name="name">[C+L] manges</field>
+        <field name="name">[C+L] manages</field>
         <field name="name_inverse">[L+C] is managed by</field>
         <field name="contact_type_left">c</field>
         <field name="contact_type_right">fsm-location</field>
     </record>
     <record id="rel_type_manages_per" model="res.partner.relation.type">
-        <field name="name">[P+L] manges</field>
+        <field name="name">[P+L] manages</field>
         <field name="name_inverse">[L+P] is managed by</field>
         <field name="contact_type_left">p</field>
         <field name="contact_type_right">fsm-location</field>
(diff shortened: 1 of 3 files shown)


fieldservice_partner_multi_relation/data/demo.xml
```diff
@@ -2,140 +2,140 @@
...
+        <field name="name_inverse">[L+C] is owned by</field>
         <field name="contact_type_left">c</field>
         <field name="contact_type_right">fsm-location</field>
     </record>
 
     <!-- manages  -->
     <record id="rel_type_manages_loc" model="res.partner.relation.type">
-        <field name="name">manges</field>
-        <field name="name_inverse">is managed by</field>
+        <field name="name">[L+L] manges</field>
```
max3903:
```suggestion
        <field name="name">[L+L] manages</field>
```

fieldservice_partner_multi_relation/data/demo.xml
```diff
@@ -2,140 +2,140 @@
...
+        <field name="name_inverse">[L+L] is managed by</field>
         <field name="contact_type_left">fsm-location</field>
         <field name="contact_type_right">fsm-location</field>
     </record>
     <record id="rel_type_manages_org" model="res.partner.relation.type">
-        <field name="name">manges</field>
-        <field name="name_inverse">is managed by</field>
+        <field name="name">[C+L] manges</field>
```
max3903:
```suggestion
        <field name="name">[C+L] manages</field>
```

fieldservice_partner_multi_relation/data/demo.xml
```diff
@@ -2,140 +2,140 @@
...
+        <field name="name_inverse">[L+C] is managed by</field>
         <field name="contact_type_left">c</field>
         <field name="contact_type_right">fsm-location</field>
     </record>
     <record id="rel_type_manages_per" model="res.partner.relation.type">
-        <field name="name">manges</field>
-        <field name="name_inverse">is managed by</field>
+        <field name="name">[P+L] manges</field>
```
max3903:
```suggestion
        <field name="name">[P+L] manages</field>
```

patrickrwilson (approved):
Tested and no longer see the error and the relationship gets created.

max3903 (approved):

sbidoul:
ocabot is watching you ;)

Issue #220, opened 2019-05-24
Order  : Description value is deleted everytime we change the location field.


rdc02271:
Hello,

Steps to reproduce problem:
1) Create a new Field Service Order
2) Write something in the Description field (fsm.order.description)
3) Select a Location (fsm.order.location_id)
4) The Description content disappears.

Thank you,
Jorge C

rdc02271:
Hello,
Just to confirm: the fsm.order.description is the field where we should write the malfunction of the equipment, correct? We would write on this field the problem description, something like, "The oven doesn't heat up". Right?

I am asking this because I was looking at the https://github.com/OCA/field-service/blob/12.0/fieldservice/models/fsm_order.py when I noticed the "Description" field is being cleared everytime the method "copy_notes" is called.

Shouldn't that method change some other field? 
In my opinion the field "Description" should be used solely to write down the problem.

Thank you,
Jorge

max3903:
Ping @patrickrwilson

patrickrwilson:
I believe the on-change event on the location field is overwriting the description with the directions from the location when the location changes. The original thought behind this is that each location may have specific directions the field worker follows each time a order is created for that location (could contain things like "Needs 24hr notice before entering" or "Front Gate Pin: 1234"). I agree that the description is probably not where this information should be set.

Suggested Improvement: Add a field called location_directions and place it just above the instructions field in the instructions tab. Then this new field is what gets updated when the location changes and the description isn't touched, using the on-change event.

wolfhall:
I like the idea to display location instructions on the order in addition to the order instructions. 
I would probably post it below the order instructions, if few locations have instructions and above if most locations have a description.

PR #243, opened 2019-06-20
[IMP] fieldservice: Add calendar view + directions to location

osi-scampbell:
Add Calendar view to FSM Orders

Added Location Directions under Instructions Tab

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -61,6 +61,7 @@ class FSMOrder(models.Model):
                                   track_visibility='always')
     location_id = fields.Many2one('fsm.location', string='Location',
                                   index=True, required=True)
+    location_directions = fields.Char(string='Location Directions')
     request_early = fields.Datetime(string='Earliest Request Date',
                                     default=datetime.now())
     request_late = fields.Datetime(string='Latest Request Date',
@@ -107,6 +108,8 @@ class FSMOrder(models.Model):
     # Planning
     person_id = fields.Many2one('fsm.person', string='Assigned To',
                                 index=True)
+    person_phone = fields.Char(related="person_id.phone",
+                               string="Worker Phone")
     route_id = fields.Many2one('fsm.route', string='Route', index=True)
     scheduled_date_start = fields.Datetime(string='Scheduled Start (ETA)')
     scheduled_duration = fields.Float(string='Scheduled duration',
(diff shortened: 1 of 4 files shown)


max3903 (approved):

patrickrwilson:
@osi-scampbell I installed but don't see the calendar view. I went to dashboard and then orders:
image

osi-scampbell:
Currently Fieldservice_geoengine overwrites the action so you do not see the calendar view. Will make a PR changing fieldservice_geoengine to include calendar as well.

patrickrwilson:
@osi-scampbell the team would like the following fields in this order:
Order #
Stage
Location
Location Phone (if possible)
Contact
Assigned To
Assigned To Phone (if possible)
City
State

I'm not sure if the phone numbers are possible, or if two fields can be on one line, and if not, more fields make the calendar look wonky, so see what you can do and how it looks.

fieldservice/models/fsm_order.py
```diff
@@ -141,12 +142,12 @@ def _onchange_customer_id_location(self):
...
     street2 = fields.Char(related="location_id.street2")
     zip = fields.Char(related="location_id.zip")
-    city = fields.Char(related="location_id.city")
+    city = fields.Char(related="location_id.city", string="City")
     state_name = fields.Char(related="location_id.state_id.name",
                              string='State', ondelete='restrict')
     country_name = fields.Char(related="location_id.country_id.name",
                                string='Country', ondelete='restrict')
-    phone = fields.Char(related="location_id.phone")
+    phone = fields.Char(related="location_id.phone" ,string="Location Phone:")
```
max3903:
```suggestion
    phone = fields.Char(related="location_id.phone" ,string="Location Phone")
```

patrickrwilson (approved):
Looks good, the views get truncated on the calendar if the appointment isn't long enough which is fine.

osi-scampbell:
@max3903 @patrickrwilson I added Location Direction field for the ticket made about this issue: https://github.com/OCA/field-service/issues/220

patrickrwilson:
@osi-scampbell Field looks good, i couldn't figure why i couldn't add a comment in the code but i think the copy_notes just needs to be updated to update that new field instead of the directions. that way your field populates instead of directions, unless there is more to it and you were just prepping this for that change?

osi-scampbell:
@max3903 I'm getting 404 Error on Runbot after my last commit?

max3903:
@osi-scampbell Try removing the sequence completely.

osi-scampbell:
@max3903 Runbot appears to be back online, only 1 warning left, 'Transient module states were reset' in \fieldservice_geoengine\views\fsm_order.xml. It throws a Key Error for 'fieldservice_geoengine.field_fsm_order__shape'

PR #245, opened 2019-06-20
[IMP] fieldservice: onchange_scheduled_date_end

osi-scampbell:
Scheduled_date_start, scheduled_date_end, and scheduled_duration were call computed together whenever one was edited. This is too restrictive, this PR sets scheduled_date_end when the scheduled_duration or scheduled_date_start is changed however if the user wants to edit the start or end date they can.

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -182,23 +182,23 @@ class FSMOrder(models.Model):
                          'request_early': str(req_date)})
         return super(FSMOrder, self).create(vals)
 
+    @api.onchange('scheduled_duration')
+    def _onchange_scheduled_duration(self):
+        if self.scheduled_date_start:
+            self.scheduled_date_end = fields.Datetime.\
+                from_string((self.scheduled_date_start) +
+                            timedelta(hours=self.scheduled_duration))
+
+    @api.onchange('scheduled_date_start')
+    def _onchange_scheduled_date_start(self):
+        if self.scheduled_date_start:
+            if self.scheduled_duration:
+                self.scheduled_date_end = fields.Datetime.\
+                    from_string((self.scheduled_date_start) +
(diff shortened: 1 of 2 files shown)


max3903:
@osi-scampbell How do you make sure that `start + duration = end`?

osi-scampbell:
@max3903 Whenever the user changes start or duration, end is calculated correctly. This was so the user could set a different end date if they wanted to. The description is on this ticket: https://pm.opensourceintegrators.com/web#id=654&view_type=form&model=helpdesk.ticket&action=334&active_id=8&menu_id=218. Patrick said it was something they would need, if this feels like more of pavlov change then I can move this to pavlov_media_fielservice

max3903:
@osi-scampbell Yes, let's move this there.

max3903 (approved):

osi-scampbell:
@max3903 I talked to Patrick and I think this solution would be easier https://github.com/pavlovmedia/pavlov-odoo/pull/226
Can we close this PR?

wolfhall:
> @max3903 Whenever the user changes start or duration, end is calculated correctly. This was so the user could set a different end date if they wanted to. The description is on this ticket: https://pm.opensourceintegrators.com/web#id=654&view_type=form&model=helpdesk.ticket&action=334&active_id=8&menu_id=218. Patrick said it was something they would need, if this feels like more of pavlov change then I can move this to pavlov_media_fielservice

wolfhall:
Most services will work with the current behaviour, where work is not interrupted or spans over multiple days.
However, here is a compromise that will work for most situations: 
1) Add a flag  "auto-calculate-end-time".  By default this is set to true
2) If the flag is set, the behaviour is exactly as it is now - the end time is calculated.
3) if it is not set, the end date will become editable and duration and end time are not calculated (e.g. duration can have 16 hrs over a week period)

PR #248, opened 2019-06-26
[IMP] fieldservice_account: FSM_Account Bill to Contact/Location, Check Acc Stage Move, Timesheet Permissions

osi-scampbell:
Add Boolean to bill contact or location, do not move stages until accounting is filled out, allow fsm dispatch/manager to add timesheet entries.

This PR also fixes an error where timesheet entries were not showing up under the timsheet app after being made from a FSM Order

### DIFF
--- a/fieldservice_account/__manifest__.py
+++ b/fieldservice_account/__manifest__.py
@@ -17,12 +17,14 @@
     ],
     'data': [
         'data/time_products.xml',
+        'data/ir_rule.xml',
         'security/ir.model.access.csv',
         'views/account.xml',
         'views/fsm_location.xml',
         'views/fsm_order.xml',
         'views/fsm_person.xml',
         'views/account_invoice_view.xml',
+        'views/hr_timesheet.xml'
     ],
     'license': 'AGPL-3',
     'development_status': 'Beta',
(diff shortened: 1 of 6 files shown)


max3903 (approved):

patrickrwilson:
Everything seems to work but i'm having trouble with the account confirm button on an order with its assigned to a vendor. I might not have my accounting setup correctly but i'm getting the following error...
The operation cannot be completed, probably due to the following:
- deletion: you may be trying to delete a record while other records still reference it
- creation/update: a mandatory field is not correctly set

[object with reference: journal_id - journal.id]

osi-scampbell:
@patrickrwilson The Accounting is probably the hardest part of Odoo (at least for me). Go to Accounting > Configurations > Journals and create a Journal. Under Advanced Settings for that Journal you can add "Account Types Allowed", add the Receivable and Payable tags. Be sure that the accounting for the product on the analytic_account_line is set, then add those Receivable and Payable accounts to the journal under "Accounts Allowed". That should do the trick.

patrickrwilson (approved):

patrickrwilson (approved):

PR #249, opened 2019-06-26
[IMP] fieldservice_sale: Add Sales Person

osi-scampbell:
Adds Sales Person to FSM Sale


max3903 (approved):

patrickrwilson (changes requested):
salesperson is already on the related partner record so there is no need to create a new field for this. Just needs added to the form. (it's labeled as salesperson but actual field is user_id)

PR #250, opened 2019-06-26
[IMP] fieldservice: Group By Error, Stage Change Check, show Scheduled, My Orders

osi-scampbell:
This PR fixes the error when trying to group by territory, branch, region, or district. It also adds checks to FSM Orders to make sure necessary fields are in place before moving stages, and allows users to skip to scheduled if the proper fields are filled out. Also adds My Orders filter for FSM Orders

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -131,13 +131,17 @@ class FSMOrder(models.Model):
 
     # Location
     territory_id = fields.Many2one('fsm.territory', string="Territory",
-                                   related='location_id.territory_id')
+                                   related='location_id.territory_id',
+                                   store=True)
     branch_id = fields.Many2one('fsm.branch', string='Branch',
-                                related='location_id.branch_id')
+                                related='location_id.branch_id',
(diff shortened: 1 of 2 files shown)


max3903 (approved):

fieldservice/models/fsm_order.py
```diff
@@ -225,18 +232,32 @@ def action_assign(self):
...
         return self.write({'stage_id': self.env.ref(
             'fieldservice.fsm_stage_started').id})
 
     def action_complete(self):
+        if not self.date_end:
+            raise ValidationError(_("Cannot move to Complete " +
+                                    "until 'Actual End' is filled in"))
+        if not self.resolution:
+            raise ValidationError(_("Cannot move to Complete " +
+                                    "until 'Resolution' is filled in"))
```
max3903, line 260: @osi-scampbell This could be grouped into one if statement. If the user do not fill in those 2 fields, he will get the error message, fix the first field, try again and get the error message for the second field...

patrickrwilson (approved):
Tested and works

PR #259, opened 2019-07-03
[FIX] fieldservice: Documentation syntax

max3903:


fieldservice/readme/CONFIGURE.rst
```diff
@@ -25,34 +25,32 @@ You need to add attribute mention below with the tag <timeline> as base element.
...
+
 You can manage designated areas or locales for your field service workers,
 salesmen, and other resources. For example, salesmen may serve a particular
 Territory. There may be multiple Territories served by a single Branch office
 location. Multiple Branches are managed within a District and these Districts
 are managed under an encompassing Region.
 
 Setup a Territory
-~~~~~~~~~~~~~~~~~
+-----------------
```
sbidoul, line 38: @max3903 the readme generator only supports tilde (~) as title underline. Other may conflict with other title levels, now or in the future.

PR #261, opened 2019-07-03
[IMP] fieldservice_account: Autopopulate description

osi-scampbell:
Autopopulate Timesheet entry decription like Contractor Costs

Only show price_unit on contractor_cost_ids to fsm managers


fieldservice_account/models/analytic_account.py
```diff
@@ -21,3 +21,8 @@ def create(self, vals):
                 raise ValidationError(_("No analytic account set "
                                         "on the order's Location."))
         return super(AccountAnalyticLine, self).create(vals)
+
+    @api.onchange('product_id')
+    def oncahnge_produc_id(self):
```
max3903:
```suggestion
    def onchange_product_id(self):
```

max3903 (approved):

max3903 (approved):

PR #238, opened 2019-06-14
[FIX] fieldservice_stock: Move Direction to Stock Request

osi-scampbell:
This PR we take out the direction field because it is being added to OCA/stock-logistics-warehouse/stock_request here: 
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/675
- [x] https://github.com/OCA/stock-logistics-warehouse/pull/606

PR #262, opened 2019-07-03
[WIP][IMP] Remove Direction from FSM Stock

osi-scampbell:
Since Stock Request now have direction added to them in OCA/stock-logistics-warehouse here https://github.com/OCA/stock-logistics-warehouse/pull/605/ there is no need to add it again in fieldservice_stock


max3903:
Duplicate of #238

Issue #277, opened 2019-07-12
[12.0][BUG] fieldservice_purchase


kongrattapong:
**Step:**

- Field Service -> Master Date -> Workers
- Create new 
- and then in worker view I already have 1 Pricelist that I never select.
![Wood Corner - Odoo - Google Chrome_014](https://user-images.githubusercontent.com/52144935/61117108-0a993680-a4c0-11e9-8b0b-fcfdf66afee6.png)

Issue #272, opened 2019-07-07
[12.0][BUG] fieldservice_account, total_cost raise error by some employee in timesheet


kittiu:
During writing test script I found following error when print order.total_cost
```
  File "/home/kittiu/ecosoft-odoo/field-service/fieldservice_account/models/fsm_order.py", line 52, in _compute_total_cost
    order.total_cost += line.unit_amount * rate
UnboundLocalError: local variable 'rate' referenced before assignment
```
Currently, we don't see the error on Work Order screen because,
1. User may choose the employee with user_id, AND
2. field "total_cost" is not show in the view, and so the method _compute_total_cost() is not called.

**Steps to Error**

1. First, we need to modify view **fieldservice_account.fsm_order_form_accounting** to also show **total_cost**
2. Create new Work Order, and select employee that has no user_id, i.e., Abigail Peterson
3. Save it, and you will see the error.

![Selection_145](https://user-images.githubusercontent.com/1973598/60766757-a568ce00-a0d8-11e9-8ddc-ffb641fa228d.png)

![Selection_146](https://user-images.githubusercontent.com/1973598/60766759-aac61880-a0d8-11e9-9b5c-6591cbc21206.png)

Issue #266, opened 2019-07-05
[12.0][BUG] set Order's Scheduled Start (ETA) to false, throw error


kittiu:
Steps:

* New Order -> Save, Scheduled Start (ETA) will be set automatically.
* Edit and set  Scheduled Start (ETA) to nothing -> Save
* Error

![image](https://user-images.githubusercontent.com/1973598/60717834-3448df80-9f4d-11e9-9420-77353cc4967a.png)

Issue #265, opened 2019-07-05
[12.0][BUG] Advance search for Linked Location throw error


kittiu:
Steps:
- Open Workers and do advance search for linked location
- Error as following,

![image](https://user-images.githubusercontent.com/1973598/60699663-22504800-9f1f-11e9-92d1-995dbd0d348c.png)

![image](https://user-images.githubusercontent.com/1973598/60699658-1ebcc100-9f1f-11e9-848b-9bfde228ddc3.png)

PR #289, opened 2019-07-18
[IMP] FSM Improvements (Bugs: #265, #272, #263, #277, #266)

osi-scampbell:
- Bug #265 Linked Location Search Error - Query needed to be adjuested to accept strings.

- Bug #272 Rate Not Declared - Declared rate = 0 at start of method. If we never find a rate then the total charges should be equal to 0

- Bug #263 equipment_count Wrong - During migration it appeared code was added that overwrote the original equipment_count calculation. This code has been removed.

- Bug #277 Worker Pricelists Computed Wrong - Computations were made with the wrong variable. Now referencing the name on the pricelist instead of the pricelist ID.

- Bug #266 ETA False Error - Only calculating scheduled_date_end if scheduled_date_start is not equal to False

### DIFF
--- a/fieldservice_account/models/fsm_order.py
+++ b/fieldservice_account/models/fsm_order.py
@@ -46,6 +46,7 @@ class FSMOrder(models.Model):
     def _compute_total_cost(self):
         for order in self:
             order.total_cost = 0.0
+            rate = 0
             for line in order.employee_timesheet_ids:
                 for emp in line.user_id.employee_ids:
                     rate = emp.timesheet_cost
--- a/fieldservice_account/models/fsm_order.py
+++ b/fieldservice_account/models/fsm_order.py
@@ -48,9 +48,7 @@ class FSMOrder(models.Model):
             order.total_cost = 0.0
             rate = 0
             for line in order.employee_timesheet_ids:
-                for emp in line.user_id.employee_ids:
-                    rate = emp.timesheet_cost
-                    continue
+                rate = line.employee_id.timesheet_cost
                 order.total_cost += line.unit_amount * rate
             for cost in order.contractor_cost_ids:
                 order.total_cost += cost.price_unit * cost.quantity
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -190,8 +190,7 @@ class FSMLocation(models.Model):
                 child_locs = self.env['fsm.location'].\
                     search([('fsm_parent_id', '=', child.id)])
                 equip = self.env['fsm.equipment'].\
-                    search_count([('location_id',
-                                   '=', child.id)])
+                    search_count([('location_id', '=', child.id)])
             if child_locs:
                 for loc in child_locs:
                     equip += loc.comp_count(0, 1, loc)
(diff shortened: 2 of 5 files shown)


kittiu:
@osi-scampbell seem that you are not base of latest fieldservice? I found it still depend on base_geoengine.
When I install this, I stuck with this error,
image

Then I do install the postgis, still got following error in install,
image

Would you rebase on latest fieldservice? Thanks!

kittiu:
I merged this PR into my github OCA, and pass following tests,

- [x] Bug #265 Linked Location Search Error - Query needed to be adjuested to accept strings.
- [x] Bug #272 Rate Not Declared - Declared rate = 0 at start of method. If we never find a rate then the total charges should be equal to 0
- [x] Bug #263 equipment_count Wrong - During migration it appeared code was added that overwrote the original equipment_count calculation. This code has been removed.
- [x] Bug #277 Worker Pricelists Computed Wrong - Computations were made with the wrong variable. Now referencing the name on the pricelist instead of the pricelist ID.
- [x] Bug #266 ETA False Error - Only calculating scheduled_date_end if scheduled_date_start is not equal to False

fieldservice_account/models/fsm_order.py
```diff
@@ -46,6 +46,7 @@ def _compute_employee(self):
     def _compute_total_cost(self):
         for order in self:
             order.total_cost = 0.0
+            rate = 0
```
kittiu, line 49: The rate is added and there is no error message. But still, because of this
```
for emp in line.user_id.employee_ids
```
If the line with selected employee_id that not have user_id, the total_cost will still be 0.0 (even employee have timesheet rates)

I am not sure this is what is intended? But for me, when seeing with my eye it not make sense.

kittiu:
Done review.

fieldservice_account/models/fsm_order.py
osi-scampbell: @kittiu I agree here that this method seems unnecessary, we can just get the timesheet_cost from the employee_id on the line. Made changes in my last commit.

osi-scampbell:
> @osi-scampbell seem that you are not base of latest fieldservice? I found it still depend on base_geoengine.
> When I install this, I stuck with this error,
> image
> 
> Then I do install the postgis, still got following error in install,
> image
> 
> Would you rebase on latest fieldservice? Thanks!

@kittiu The latest version of fieldservice does depend on base_geoengine as seen here on branch 12.0 for field-service: https://github.com/OCA/field-service/blob/12.0/fieldservice/__manifest__.py

kittiu:
> > @osi-scampbell seem that you are not base of latest fieldservice? I found it still depend on base_geoengine.
> > When I install this, I stuck with this error,
> > image
> > Then I do install the postgis, still got following error in install,
> > image
> > Would you rebase on latest fieldservice? Thanks!
> 
> @kittiu The latest version of fieldservice does depend on base_geoengine as seen here on branch 12.0 for field-service: https://github.com/OCA/field-service/blob/12.0/fieldservice/__manifest__.py

My mistake. Now, I checked all passed.

max3903 (approved):

kittiu (approved):

PR #296, opened 2019-07-25
[FIX] fieldservice_account: Wrong journals

max3903:
If your user has access to multiple companies (children ones), you can end up with the wrong journals...


brian10048 (approved):
Changes LGTM, no test

kittiu (approved):
LGTM

PR #303, opened 2019-08-01
[12.0][IMP] fieldservice_agreement, add unit test

kittiu:
Done  with the test script, with some remarks as follows.

### DIFF
--- a/fieldservice_agreement/tests/__init__.py
+++ b/fieldservice_agreement/tests/__init__.py
@@ -1,0 +1,4 @@
+# Copyright (C) 2019 - TODAY, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import test_fsm_agreement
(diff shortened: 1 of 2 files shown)


kittiu:
Test will error by this line,
https://github.com/OCA/field-service/blob/12.0/fieldservice_agreement/models/fsm_person.py#L20-L21
https://github.com/OCA/field-service/blob/12.0/fieldservice_agreement/models/fsm_person.py#L26-L27
The external id will has space when ref() in code

kittiu:
I also have question about following code, while it works but should we 1) Use for loop 2) Use [('partner_id.name', '=', self.name)]

https://github.com/OCA/field-service/blob/12.0/fieldservice_agreement/models/fsm_person.py#L13-L15

kittiu:
I also have question, looking into the code in **fsm_person.py**, it looks like the field `partner_id` in agreement should be **Worker** and not Customer ?

Just wonder, but no objection.

max3903:
@kittiu The agreement is linked to `partner_id` to apply to customer, vendor and worker.

kittiu:
@max3903 let me know if this PR need any changes.

max3903 (approved):

osi-scampbell (approved):

Issue #297, opened 2019-07-26
[ADD] fieldservice_stock_account


max3903:
If inventory items were delivered to a customer location, the delivery will generate an accounting entry in the stock output account. The balance of that account will keep increasing as those inventory items will not get invoiced.

This module allows you to automatically adds inventory items delivered as part of the field service order to a customer invoice to account for their value as a cost of goods (or service) sold.

# Usage

* Go to Field Service
* Create an order for a customer location (type of the inventory location must be 'Customer')
* Add inventory items and process them. Their delivery will generate an accounting entry.
* Complete the field service order
* Add some contractor costs or timesheet entries depending on the worker (vendor or employee)
* In the accounting tab, click on "Create invoice". The delivered inventory items will be added to the customer invoice with a 0$ price unit and they will not show up in the PDF report of the invoice.
  * Go to Invoicing > Customers > Invoices
  * Validate the invoice. 
  * Check the accounting entry to make sure the value of the inventory items is accounted as  a cost of goods/services sold.
* Repeat the process but in the accounting tab, click on "No invoice"
  * Go to Invoicing > Customers > Invoices
  * Open the invoice related to the field service order
  * Check that the invoice has a 0$ total and is paid
  * Check the accounting entry to make sure the value of the inventory items is accounted as a cost of goods/services sold.

# Technical Specifications

Dependencies: fieldservice_account, fieldservice_stock, account_invoice_report_hide_line

- [x] https://github.com/OCA/account-invoice-reporting/issues/114

## fsm.order

- [x] Extend account_create_invoice(), if the FSM location is linked to an inventory location of type "Customer":
  * add all the stock requests (stock_request_ids) in done state to the invoice with:
    * show_in_report = False
    * quantity = qty_done
    * price_unit = 0
- [x] Extend account_no_invoice(), if the FSM location is linked to an inventory location of type "Customer":
  * create a customer invoice with the same header info as in account_create_invoice(). No invoice line from the contractor costs.
  * add all the stock requests in done state as invoice lines and:
    * show_in_report = False
    * quantity = qty_done
    * price_unit = 0
  * validate the invoice --> because the total is 0 $, the invoice is paid

PR #304, opened 2019-08-01
[12.0][ADD] fieldservice_stock_account

bodedra:
Issue Reference: https://github.com/OCA/field-service/issues/297

Depends on :
- [x] #238 
- [x] #305

CC @max3903


bodedra:
@osi-scampbell @brian10048 are you guys available for review these ?

max3903 (approved):

brian10048:
@max3903 @bodedra 

I might not have configured something properly or I am misunderstanding the intended purpose.

In the issue reference #297, after validating the invoice the expectation is 

> Check the accounting entry to make sure the value of the inventory items is accounted as a cost of goods/services sold.

The journal entry I get shows the product as a label with its income (sales) account rather than an expense (cost of goods) account.  The value is also 0.  How are we to account for the cost using the income account for that product and a value of 0?  For the invoice purpose I understand we aren't charging specifically for the product so it needs to be 0 on the invoice, but how will this allow tracking its cost in the invoice's journal entry?

max3903:
@brian10048 Did you activate the anglo-saxon accounting?

brian10048:
@max3903 
I installed the US Chart of Accounts module, so this means Anglo-Saxon is activated, correct?

After looking further, would I need to be using a product with 'automated' inventory valuation, not 'manual'?

max3903:
@brian10048 Yes and yes

brian10048 (approved):
Tested and working as intended

PR #315, opened 2019-08-20
[12.0][FIX]fieldservice_recurring - create all recurring orders at confirmation

### LINKED ISSUE #312: [12.0] fieldservice_recurring don't create order on confirm
I guess it's the same for previous versions, but.

Create a frequency like : Every Monday, weekly.
Create a recurring order with this frequency on Friday, and confirm it. 
An order will be created on Friday (datetime.now()).

The date of this order is not compliant with the frequency defined.

I suggest to remove _create_order from action_start and let the cron create the orders.

```py
    def action_start(self):
        for rec in self:
            if not rec.start_date:
                rec.start_date = datetime.now()
            rec._create_order(date=rec.start_date)
            rec.write({'state': 'progress'}
```
https://github.com/OCA/field-service/blob/12.0/fieldservice_recurring/models/fsm_recurring.py#L101

hparfr:
Fix #312

### DIFF
--- a/fieldservice_recurring/models/fsm_recurring.py
+++ b/fieldservice_recurring/models/fsm_recurring.py
@@ -98,8 +98,8 @@ class FSMRecurringOrder(models.Model):
         for rec in self:
             if not rec.start_date:
                 rec.start_date = datetime.now()
-            rec._create_order(date=rec.start_date)
             rec.write({'state': 'progress'})
+            rec._generate_orders()
 
     @api.multi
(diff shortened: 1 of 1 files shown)


hparfr:
@brian10048

brian10048 (approved):
Tested and functioning as intended

max3903 (approved):

osi-scampbell (approved):

PR #320, opened 2019-08-23
[12.0][FIX]fieldservice calc scheduled dates and duration

hparfr:
Before this PR:
when an order is created with schedule_date_start + duration, the schedule_date_end is not calculated.

After :
the behaviour is consistent across create() and write()


max3903 (approved):

brian10048 (approved):
Code review only, not tested.  Thank you!

bodedra (approved):

PR #326, opened 2019-09-05
[ADD] fsm_location_builder

osi-scampbell:
This PR adds fsm_location_builder, a tool used to create multiple fsm_locations in a short period of time. This way, when a user has a building with 200 units on 4 floors in 2 buildings, the user can use this tool instead of manually creating all 2000 units manually.

#128

Issue #129, opened 2019-02-04
[ADD] fieldservice_stock_location_builder


max3903:
License: AGPL-3
Dependencies: fieldservice_location_builder, fieldservice_stock
Auto_install: True

# Module description

This module sets the inventory location on the sub-locations when running the location builder wizard.

# Usage

See https://github.com/OCA/field-service/tree/11.0/fieldservice_location_builder#id1

bodedra:
@max3903 Usage link is dead.

max3903:
@bodedra It is meant to point to this content: https://github.com/OCA/field-service/issues/128

PR #327, opened 2019-09-05
[ADD] fieldservice_stock_location_builder

osi-scampbell:
This PR overrides the create_sub_locations() method in FSMLocationBuilderWizard so all sublocations have the same inventory location as the parent location.

Depends on #326

#129


patrickrwilson (changes requested):
Works well, will approve once all green.

layadelcarmen:
I have made a functional review of this module. The wizard creates the new locations with the parent inventory location. Wizard lines needs some validation, admit empty level name and odoo throw an error .`    name + spacer + str(lev_id),
TypeError: unsupported operand type(s) for +: 'bool' and 'str'`

fieldservice_stock_location_builder/wizard/fieldservice_location_builder_wizard.py
```diff
@@ -0,0 +1,33 @@
...
+                        level_ids[num].name + spacer + str(lev_id),
+                        'owner_id': location.owner_id.id,
+                        'customer_id': location.customer_id.id,
+                        'fsm_parent_id': parent.id,
+                        'inventory_location_id': location.
+                        inventory_location_id.id}
+                new_location = self.env['fsm.location'].create(vals)
+                if num < levels:
+                    build_location(new_location, num + 1)
+        build_location(location, 0)
```
max3903: @osi-scampbell I would add a method `prepare_fsm_location_values()` in `fieldservice_location_builder` so you can extend it here just by adding the inventory location to the dictionary.

max3903:
Ping @osi-scampbell

fieldservice_stock_location_builder/__manifest__.py
```diff
@@ -10,7 +10,8 @@
     'website': 'https://github.com/OCA/field-service',
     'depends': [
         'fieldservice_stock',
-        'fieldservice_location_builder'
+        'fieldservice_location_builder',
+        'fieldservice_account'
```
max3903: @osi-scampbell Why do you need `fieldservice_account`?

osi-scampbell: That is where we moved customer_id to

max3903: @osi-scampbell The `customer_id` field on `fsm_location` is added in `fieldservice_account_analytic`.

1. Let's remove the dependency here
2. In `fieldservice_account_analytic`, add a default value on the field with a method:
```
if self.parent_id:
    return self.parent_id.customer_id.id
return self.owner_id.id
```

osi-scampbell:
#474

PR #342, opened 2019-09-16
[IMP] fieldservice: Allow dispatcher to delete order in default stage

max3903:


patrickrwilson (approved):
Tested and works correctly

fieldservice/models/fsm_order.py
```diff
@@ -224,6 +224,16 @@ def write(self, vals):
                 order.customer_id = order.location_id.customer_id.id
         return res
 
+    @api.multi
+    def unlink(self):
+        for order in self:
+            if order.stage_id == order._default_stage_id():
+                res = super(FSMOrder, order).unlink()
```
hparfr, line 236: May I suggest to put the test in a seperate function instead of directly in unlink ?
```py
if order.can_unlink():
  res = super().unlink()
else:
  raise ...
```
Otherwise it's hard to change the conditions in inherited classes.

hparfr (changes requested):

hparfr (approved):
Thanks !

bodedra (approved):
LGTM :+1:

PR #348, opened 2019-09-23
[12.0] fieldservice fsm.order Fix inheritance on default=

hparfr:
We can't inherit a method defined on field.default.

Consistent with stage_id few lines above.


bodedra (approved):
LGTM :+1:

max3903 (approved):

Issue #346, opened 2019-09-19
[12.0] maintenance: Unnecessary dependence on hr


zbik2607:
I suggest removing the dependency on the hr module in module fieldservice-maintenance, is deprecated.

In addition, dependency on the fieldservice_stock module also also seems debatable and redundant.

bodedra:
@zbik2607 Thanks for notice.

bodedra:
@max3903 We may close this.

PR #241, opened 2019-06-18
[IMP] fieldservice: Support multiple equipments on service request

osi-scampbell:
This PR allows users to add multiple Equipment on a Service Request. It is important to note that only 1 Equipment can be used on a Repair Order or Maintenance Order. So when those are selected, we hide the Many2many field and make the user select 1 Equipment.

PR #364, opened 2019-10-10
[12.0][RM] Unnecessary dependence on hr and fieldservice_stock module

bodedra:
Issue Reference: https://github.com/OCA/field-service/issues/346

### DIFF
--- a/fieldservice_maintenance/models/fsm_order.py
+++ b/fieldservice_maintenance/models/fsm_order.py
@@ -1,8 +1,7 @@
 # Copyright (C) 2018 - TODAY, Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
-# from odoo import api, fields, models
+from odoo import api, fields, models
 
 
 class FSMOrder(models.Model):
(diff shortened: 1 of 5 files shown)


bodedra:
CC @max3903

fieldservice_maintenance/models/fsm_order.py
```diff
@@ -10,31 +10,3 @@ class FSMOrder(models.Model):
...
-    #                         equipment.maintenance_equipment_id.id,
-    #                     'category_id': equipment.category_id.id,
-    #                     'request_date': fields.Date.context_today(self),
-    #                     'maintenance_type': 'corrective',
-    #                     'maintenance_team_id': team_id,
-    #                     'schedule_date': order.request_early,
-    #                     'description': order.description
-    #                 })
-    #             order.request_id = request_id
-    #     return order
```
max3903: @bodedra We need to uncomment this method to create the maintenance request.

bodedra: @max3903 This method is already in comment. I just remove commented code.

If our goal to uncomment this method, then `hr` module is a valid as a dependency.

max3903: We can remove the employee_id field and the dependency on hr.

bodedra: Sure. Will do it.

fieldservice_maintenance/models/fsm_order.py
```diff
@@ -11,30 +11,30 @@ class FSMOrder(models.Model):
...
+    @api.model
+    def create(self, vals):
+        # if FSM order with type maintenance is create then
+        # create maintenance request
+        order = super(FSMOrder, self).create(vals)
+        if order.type == 'maintenance':
```
max3903: @bodedra The type is now a many2one field. We need to use the external id.

bodedra: @max3903 Yes. I first uncomment lines and correct the logic.

bodedra:
@max3903 Comment attended!

max3903 (approved):

max3903:
@zbik2607 Can you review and provide your feedbacks?

bodedra:
Are you available for review ? @brian10048 @hparfr

hparfr (commented):
Why not make use of inheritence instead of create() in create() ?

bodedra:
@hparfr IMO That's the another way of doing it but this will not harm. We are only create a maintenance request if FSM order type is maintenance.

zbik2607:
if equipment.maintenance_equipment_id is not set, then creating maintenance.request probably doesn't make sense?

bodedra:
> if equipment.maintenance_equipment_id is not set, then creating maintenance.request probably doesn't make sense?

Yes. Agree. I will improve this part because 'maintenance_equipment_id' field is not required.

Good catch.

bodedra:
@zbik2607 Comment addressed!

zbik2607:
I suggest deleting also dependency on fieldservice_stock - it is redundant too. 
Also replace lines in fsm_equipemt.py:

            'serial_no':
                vals['lot_id'] and
                self.env['stock.production.lot'].browse(vals['lot_id']).name
                or False

by:

            'serial_no':
                vals.get('lot_id', False) and
                self.env['stock.production.lot'].browse(vals['lot_id']).name

and remove view: fsm_equipment_form_view_model, move field model to fsm_equipment_form_view_maintenance

zbik2607:
Sorry, for small fix  in updated my last answer:

"and remove view: fsm_equipment_form_view_model, move field model to fsm_equipment_form_view_maintenance"

zbik2607:
Its OK :)

bodedra:
@zbik2607 That's great. Thanks for your review. You can approve it.

zbik2607:
Rather, it should be: vals.get ('lot_id', False)
instead of vals ['lot_id']
because lot_id may be absent

bodedra:
@zbik2607 Comment addressed!

zbik2607:
Everything seems to be correct now.

bodedra:
> Everything seems to be correct now.

Sounds good. Thank you. You can give your official approval vote from the right hand side, file changes. :)

zbik2607 (approved):
Everything seems to be correct now.

brian10048:
Do we know why @osi-scampbell commented the create method in #241?

@bodedra It may be worth improving fieldservice_repair as well? https://github.com/OCA/field-service/blob/ba82cadc01ecc35194447bcff595f27efd86ee45/fieldservice_repair/models/fsm_order.py#L15

bodedra:
@brian10048 I don't the reason. @osi-scampbell would you please help us to here ?

Sure. Let's create a separate issue for improving fieldservice_repair module.

max3903:
> Do we know why @osi-scampbell commented the create method in #241?

It was not working well enough.

> @bodedra It may be worth improving fieldservice_repair as well?
> 
> https://github.com/OCA/field-service/blob/ba82cadc01ecc35194447bcff595f27efd86ee45/fieldservice_repair/models/fsm_order.py#L15

Yes

zbik2607:
I see other errors, but it's probably better to approve and create new issues. For example:  https://github.com/OCA/field-service/issues/370

hparfr (approved):

max3903:
@bodedra Can you fix #370 as it is related to the same module?

bodedra:
@max3903 suggested code from @zbik2607 is added.

bodedra:
@max3903 This is ready for merge.

PR #333, opened 2019-09-06
[12.0]fieldservice - Remove most of the stages and its validation logic

hparfr:
See issue #309.

This PR keeps only 3 stages: new, complete and canceled.
The other stages are moved PR : #334

PR #368, opened 2019-10-14
[FIX] Unassigned Unscheduled Order Filters

brian10048:
In #333 most of the stages were removed.

The following search filters were also moved, but do not use the removed stages for any logic.  These filters are used in base fieldservice module for the Teams dashboard and are also useful as filters in the search view.

This PR is to move them back into base fieldservice


max3903 (approved):

bodedra (approved):
Make sense :+1:

hparfr (approved):

PR #376, opened 2019-10-28
[IMP] fieldservice: Add Owned Locations to Partner

patrickrwilson:
Add Owned Locations field to partner view in the Field Service page.
This gives visibility to show all parent locations owned by the partner.

### DIFF
--- a/fieldservice/models/res_partner.py
+++ b/fieldservice/models/res_partner.py
@@ -11,3 +11,8 @@ class ResPartner(models.Model):
     fsm_person = fields.Boolean('Is a FS Worker')
     service_location_id = fields.Many2one('fsm.location',
                                           string='Primary Service Location')
+    owned_location_ids = fields.One2many('fsm.location',
+                                         'owner_id',
+                                         string='Owned Locations',
+                                         domain=[('fsm_parent_id', '=', False)]
+                                         )
--- a/fieldservice/models/res_partner.py
+++ b/fieldservice/models/res_partner.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2018 - TODAY, Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ResPartner(models.Model):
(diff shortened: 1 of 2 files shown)


patrickrwilson:
Ping @max3903 @osi-scampbell

max3903:
@patrickrwilson What do you think of using a smart button for that? A company may own a lot of locations...

patrickrwilson:
> @patrickrwilson What do you think of using a smart button for that? A company may own a lot of locations...

Good idea, I'll make that change.

hparfr:
Why not using the partner address type  ?
https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_partner.py#L182

patrickrwilson:
> Why not using the partner address type ?
> https://github.com/odoo/odoo/blob/12.0/odoo/addons/base/models/res_partner.py#L182

@hparfr , i'm not sure what you mean. Where are you suggesting I use the address type?

hparfr:
I suggest to use the address_type instead of adding a new field.

patrickrwilson:
> I suggest to use the address_type instead of adding a new field.

Sorry, I'm struggling to understood. On the FSM Location, there is a owner_id field. Although you could filter the locations owned by a partner it's a pain to do so, therefore this improvement adds a smart button to the partner form to show all locations owned by that partner. Maybe i'm not sure how the address type would be used to make that button show the owned locations.

max3903 (approved):

osi-scampbell (approved):

fieldservice/models/res_partner.py
```diff
@@ -11,3 +11,35 @@ class ResPartner(models.Model):
...
+    owned_location_count = fields.Integer(
+        compute='_compute_owned_location_count',
+        string="Owned Locations")
+
+    @api.multi
+    def _compute_owned_location_count(self):
+        for partner in self:
+            res = self.env['fsm.location'].search_count(
+                [('owner_id', '=', partner.id)])
+            partner.owned_location_count = res or 0
```
bodedra:
```suggestion
            partner.owned_location_count = res
```

bodedra (approved):
LGTM :+1: added non blocking comment.

hparfr:
Because you are working on this, I'm suggesting a different design:

1) remove / don't use  / prefill [fsm.location]owner_id and use instead [res.partner].parent_id
2) (optional) use partner_type to add a little logic (like create res.partner then convert-it into fsm.location).
3) keep your smart button to filter on child_ids and partner_type = fsm.location


Screenshot_2019-10-31 Ready Mat - Odoo

Screenshot_2019-10-31 Ready Mat - Odoo(1)

May be I miss the point here. But I do not see why having this owner_id is needed in the baes module fieldservice.

max3903:
@hparfr If we do that, FSM location cannot have an address type:
* Some locations can be a "Shipping address" where materials can be received
* Other may be an "Other address" because there is noone there to contact or receive materials

patrickrwilson:
@hparfr I see what you mean.  I'd like for others to chime in on your suggestion as I think the locations were not intended to be a children of partners but their own entity with the relationship as the linkage. Keep in mind that locations can change ownership often (like in a condo/apt situation) so more logic might be needed if the owner changes on the location. This particular PR was just to fulfill a need we had with how it's currently implemented but your suggestion is an interesting approach.

hparfr:
@max3903 , agree a FSM Location as partner_type may be a bit too restrictive. 
@patrickrwilson, about the changing ownership, I don't see it as problem. In all my use cases, if the owner changes, usually contacts changes as well and old history is not very relevant. Plus I guess you can change always the parent_id of a contact. 

I will try some experiments on this soon.

max3903:
@hparfr Meanwhile you test, do you mind if we merge?

hparfr (approved):

PR #378, opened 2019-10-30
[MIG] FSM Order Type Migration

osi-scampbell:
This is the migration script + demo data needed for the migration script. To execute the migration script, upgrade the fieldservice module from the backend.


hparfr:
Shoudn't it be in demo instead ? Or if data then in related modules ?

max3903:
@hparfr It should be private... ;)

PR #367, opened 2019-10-11
[FIX] Travis and Runbot errors

max3903:
... so the pot files can be generated and pushed to https://translation.odoo-community.org

PR #386, opened 2019-11-07
[FIX] fieldservice_account: required customer_id no longer breaks installation

dreispt:
Relevant for #367


max3903:
@dreispt I would keep the field required and have a hook to set the value to the same as the owner by default.

brian10048:
I was thinking it should stay required as well

max3903:
@dreispt FYI
https://github.com/OCA/field-service/pull/367/files#diff-2df1f2c4a8f6d8a31e786e8a3b39bc72R16

dreispt:
A required field imposes something to all use cases, so you should make sure you have a strong reason to do so.
I think there are valid scenarios where we don't need a Location Customer, and even think it would be useful to have the customer invoicing features disabled for those cases.

So I still think we will have a better solution keeping this field optional.

PR #387, opened 2019-11-07
[RFC][IMP][12.0] fieldservice_account

brian10048:
This PR adds the ability to create an invoice from a field service order with a fixed price and product, while still allowing a field service order to be invoiced based on the existing implementation using timesheets and/or contractor costs

### DIFF
--- a/fieldservice_account/models/fsm_order.py
+++ b/fieldservice_account/models/fsm_order.py
@@ -46,12 +46,11 @@ class FSMOrder(models.Model):
                                    string='Cost Method',
                                    required=True,
                                    default='timesheet')
-    fixed_cost = fields.Float(string='Fixed Cost')
     product_id = fields.Many2one('product.product', string='Invoice Product')
     invoice_id = fields.Many2one('account.invoice', string='Customer Invoice',
-                                 domain="[('type', '=', 'out_invoice')")
+                                 domain="[('type', '=', 'out_invoice')]")
     bill_id = fields.Many2one('account.invoice', string='Vendor Bill',
-                              domain="[('type', '=', 'in'_invoice')]")
+                              domain="[('type', '=', 'in_invoice')]")
 
     def _compute_employee(self):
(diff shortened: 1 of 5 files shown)


max3903:
@brian10048 What do you think of
- [x] adding a many2one field "Customer Invoice" to account.invoice on fsm.order
- [x] adding a many2one field "Vendor Bill" to account.invoice on fsm.order
- [x] hiding those fields if they are not set
- [x] setting those fields when the bill or invoice are created?

brian10048:
@max3903 I can handle those items.

If there is anything else you would like changed, just let me know and I'll make the revisions.

max3903:
@brian10048 Thank you! What's your opinion on #386?

fieldservice_account/models/fsm_order.py
```diff
@@ -32,31 +33,45 @@ class FSMOrder(models.Model):
...
                                   domain=[('customer', '=', True)],
                                   change_default=True,
                                   index=True,
                                   track_visibility='always')
+    cost_method = fields.Selection([('timesheet', 'Timesheets'),
+                                    ('fixed', 'Fixed Rate')],
+                                   string='Cost Method',
+                                   required=True,
+                                   default='timesheet')
+    fixed_cost = fields.Float(string='Fixed Cost')
```
max3903: @brian10048 Why not use the product below to get the costing method from the product category and the standard cost from the product?

dreispt: ... and apply Price Lists to get the amount to bill.

dreispt (commented):
LGTM, agree with Max on using Product data instead of a fixed price float.

fieldservice_account/models/fsm_order.py
brian10048: @max3903 @dreispt - That makes more sense

Your suggestions have been made.

brian10048:
Not relevant since #411

PR #388, opened 2019-11-10
[IMP] fieldservice_sale: Compute qty_delivered

brian10048:
- Completed field service orders will now update the qty_delivered field on linked SO line


fieldservice_sale/readme/USAGE.rst
```diff
@@ -4,3 +4,6 @@
 * On a Sale Order Line, select a product configured for field service orders
 * Confirm the Sale Order
 * Field Service orders linked to SO lines are created
+* When a Field Service order is completed, the quantity delivered for its
+linked sale order line will be updated to the quantity ordered, indicating
```
dreispt: Indent 3 spaces to have a RST correct multiline bullet paragraph.

brian10048: Thank you! Fixed and squashed

bodedra (approved):
Code review. LGTM :+1:

fieldservice_sale/models/sale_order_line.py
```diff
@@ -31,6 +33,26 @@ def _compute_product_updatable(self):
...
+    def _compute_qty_delivered_method(self):
+        super(SaleOrderLine, self)._compute_qty_delivered_method()
+        for line in self:
+            if not line.is_expense and line.product_id.type == 'service' \
+               and line.product_id.field_service_tracking == 'order':
+                line.qty_delivered_method = 'field'
+
+    @api.multi
+    @api.depends('fsm_order_id.stage_id')
+    def _compute_qty_delivered(self):
```
hparfr, line 47: May be dumb question, but are you sure it will not interfere with https://github.com/odoo/odoo/blob/12.0/addons/sale_stock/models/sale_order.py#L202 ?

fieldservice_sale/models/sale_order_line.py
```diff
@@ -8,6 +8,8 @@
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
+    qty_delivered_method = fields.Selection(
+        selection_add=[('field', 'Field Service Order')])
```
hparfr: 'field' is unclear. may be use field_service ?

fieldservice_sale/models/sale_order_line.py
```diff
@@ -31,6 +33,26 @@ def _compute_product_updatable(self):
...
+
+    @api.multi
+    @api.depends('fsm_order_id.stage_id')
+    def _compute_qty_delivered(self):
+        super(SaleOrderLine, self)._compute_qty_delivered()
+        lines_by_fsm = self.filtered(lambda sol:
+                                     sol.qty_delivered_method == 'field')
+        complete = self.env.ref('fieldservice.fsm_stage_completed')
+        for line in lines_by_fsm:
+            if line.fsm_order_id.stage_id == complete:
```
hparfr, line 56: initialize qty with 0 otherwise, if you change state, qty will still retain the old value.

hparfr (changes requested):
A small request about qty_delivered_method and initialization of qty

fieldservice_sale/models/sale_order_line.py
brian10048: I am not 100% positive.  But [https://github.com/odoo/odoo/blob/12.0/addons/sale_timesheet/models/sale_order.py#L142](https://github.com/odoo/odoo/blob/12.0/addons/sale_timesheet/models/sale_order.py#L142) does something very similar to what I did, so I assume there will be no issues here.

fieldservice_sale/models/sale_order_line.py
brian10048: Sounds good to me

hparfr (approved):
Thanks

Issue #393, opened 2019-11-13
[RFC] fieldservice_sale


max3903:
This change will allow to create an FSM order for each sales order and link it to the delivery order and invoice.

# product.template

- [x] Add a selection field "Service Policy" (fsm_policy) with the options:
* "No service"
* "One service per sale order"
* "One service per sale order line"

# fsm.order

- [x] Add a many2one field "Sale Order" (sale_id) to sale.order

# sale.order

- [x] When the sale order is confirmed:
* For each line
  * if the product service policy is No service, do nothing (default Odoo behavior)
  * if the product service policy is One service per sale order, create a FSM order and set it on the sale order if it's not already done
  * if the product service policy is One service per sale order line, create a FSM order for each line
* set the fsm order on the pickings created by the sale order
- [x] When an invoice is created, set the fsm_order_id field to the related FSM Order of the line (or the order)
- [x] Add a smart button to show the FSM orders related to the sale order. Only when the sale order is confirmed (same behavior as picking).

| Products Service Policy | FSM Orders on Sale Order | FSM Order on Picking | FSM Order | FSM Order on Invoice |
|--------------------------------|----------------|-----------|-----------------|-------------|
| No service | False | False |  | False |
| One service per sale order | ID1 | ID1 | ID1 | ID1
| One service per sale order line | ID2 | ID2 | ID2 | ID2
| Multiple products with One service per sale order line | ID2 and ID3 | False | 2 FSM Orders with ID2 and ID3 | 2 invoices with ID2 and ID3
| Mixed products | ID1, ID2 and ID3 | ID1 | ID1, ID2 and ID3 | 3 invoices with ID1, ID2 and ID3

brian10048:
@max3903 Can we figure out a way so that this isn't a company setting, but maybe a setting per each new sale order? I need to be able to do both methods of order creation in one company

brian10048:
Is there a way this can be accomplished using routes on the products and choosing a route on each sale order line?

I will use my company specific examples for lack of better scenarios

I have some field service order templates as follows.  Each of these templates require different resources (ex: specialized vehicle and worker skills) to complete. 

| FS Order Template |
| ------------------ |
| Lawn Fertilization Order |
| Lawn Mowing Order |
| Sprinkler System Work Order |
| Landscape Enhancement Work Order |

These are some example products I sell on sale orders

| Product  | Type | related FS Order Template | Service Policy |
| ---------| ---- | --------------------| ---------------------- |
[...]

PR #403, opened 2019-11-21
[12.0][IMP] fieldservice_sale

brian10048:
#393

Depends on:
- [x] #411


fieldservice_sale/models/sale_order.py
```diff
@@ -25,10 +25,76 @@ class SaleOrder(models.Model):
     @api.depends('order_line.product_id')
     def _compute_fsm_order_ids(self):
         for order in self:
-            order.fsm_order_ids = self.env['fsm.order'].search([
+            orders = self.env['fsm.order']
+            orders += self.env['fsm.order'].search([
```
hparfr: I think it's better to use |= because it's an union of recordsets.

fieldservice_sale/models/sale_order.py
```diff
@@ -25,10 +25,76 @@ class SaleOrder(models.Model):
...
+                """This order has been created from: <a href=
+                   # data-oe-model=sale.order data-oe-id=%d>%s</a>
+                """) % (so.id, so.name)
+            fsm_order.message_post(body=fsm_order_msg)
+            result[so.id] = fsm_order
+        return result
+
+    @api.multi
+    def _field_find_fsm_order(self):
+        """ Find the fsm_order generated by the Sale Order. If no fsm_order
```
hparfr, line 85: I think it's better to adapt the same behavior as delivery orders : when you cancel the sale, the picking is cancelled, when set to quotation then confirmed again a new picking is created.

hparfr (changes requested):
There is a lot of code duplication between sale_order and sale_order_line about creating fsm_orders.

fieldservice_sale/models/sale_order.py
brian10048: Thank you.  Updated in a recent commit

fieldservice_sale/models/sale_order.py
brian10048: That is probably a better method than this is being done here.  I agree there is a fair amount of code duplication between sale_order and sale_order_line.  While similar they are also a little different.

I originally modeled this after v11 of sale_timesheet.

max3903:
@brian10048 Nice job. Please see https://github.com/brian10048/field-service/pull/6

brian10048:
@max3903 I made the additions to copy template information to the field service order

fieldservice_sale/__manifest__.py
```diff
@@ -11,6 +11,7 @@
     'depends': [
         'fieldservice',
         'sale_management',
+        'fieldservice_account',
```
hparfr, line 14: Please don't add this dependency.

hparfr (changes requested):
Accounting should not be a dependency of this module.

fieldservice_sale/__manifest__.py
brian10048: @hparfr I agree adding the dependency adds a lot of extra logic from fieldservice_account that is not needed for this addition, but if this dependency is not used, there is no `fsm_order_id` field for `account.invoice` model which we set here from the `sale_order`

Thoughts?

max3903: @hparfr `sale_management` depends on `account`.

If a FSM product is set to be invoiced on quantity delivered, we need the ability to invoice the completed FSM order.

hparfr: Yes, but fieldservice_account take strong assumption about how invoicing should be done
and some dependencies (analytic, hr_timesheet)

max3903: How would you do it?

brian10048: Can we split fieldservice_account so that the way it does invoicing is in a separate module and keep only things like adding fsm_order_id to the account.invoice model?
I don't mind doing the work

brian10048: @max3903 @hparfr Take a look at #411 and please provide your feedback

hparfr: Thanks for this work @brian10048.
I still think billing_id (customer_id) and costs should not be in fsm_order. There is not so much stuff left in fieldservice_accounting but just the link between account_invoice_line and fsm_order

IMO fieldservice_sale should: 
- link fsm_orders to sale_orders (ok)
- choose how fsm_orders are created (ok field_service_tracking)
- create fsm_orders from sale (ok)
- compute invoiceable quantity (ok)
- prepare invoice lines (ok)

max3903: @hparfr I am fine with this. On the last 2 items (compute invoiceable quantity and prepare invoice lines), if you invoice on delivered quantity, you would invoice from the sale order, right? Not from the FSM order (like inventory items are not invoiced from the delivery order).

# fieldservice_sale

* Invoice the service from the sale order before or after the completion of the service (--> no timesheet, no cost)

# fieldservice_account

* Invoice timesheet/costs from the FSM order

# fieldservice_sale_account

* Invoice the service + timesheet/costs from either the FSM order or the sale order, before (no timesheet/cost) or after the completion of the service

hparfr: Indeed @max3903 it's like sale orders and pickings:  _compute_qty_delivered() is currently in fieldservice_sale/models/sale_order_line.py

In the actual proposal
- fieldservice_account: links a fsm_orders to invoices lines. Kind of base module.
- fieldservice_account_analytic:  add timesheet, costs, billing_id, and so on. Depends of fieldservice_account
- fieldservice_sale, create fsm orders at sale confirmation and set quantities when invoice is created from sale. Depends of fieldservice_account
- fieldservice_contract, set quantities when invoice is created from contract. Depends of fieldservice_account

max3903: :+1:

max3903: @hparfr @brian10048 See #415 where I copied `fieldservice_account` and removed the `fsm_order_id` field on `account.invoice.line`

hparfr: @max3903  have you seen https://github.com/brian10048/field-service/pull/8 ? 
I will look at #415 in few days.

brian10048: I will be able to look at all of this over the next few days as well and provide feedback on all of the above

max3903:
@brian10048 The Field Service tracking logic seems to only apply to service product. I need it to apply to stockable product as well.

I created 4 stockable products: 2 tracked by sale, 2 tracked by line and got no FSM order when confirming the sale order.

My scenario is that I am delivering stockable products (carpet or water bottle) and both are considered delivered when:
* the stock picking is transferred
* the invoice is paid
* the FSM order is completed (product is installed, customer signed)

Edit: Thinking about this, maybe we should add a column on the sale order line "Qty Serviced" and not touch the "Qty Delivered" logic.
The quantity delivered would still be computed based on the stock moves.
The quantity serviced would be computed based on the completion of the related FSM order.
It would be clear on the sale order to see whether or not an item has been delivered or not, and installed or not.
Some organizations could invoice based on the quantity delivered or based on the quantity delivered and serviced.

What do you think?

CC: @wolfhall

brian10048:
@max3903 Did you have fieldservice_sale_stock module installed?  I don't think it is in the settings page.  It should provide the functionality you mentioned needing for creating the field service order with a stockable product.  I separated it to help minimize dependencies of fieldservice_sale

I'm not opposed to a qty_serviced field for the sale order line

fieldservice_sale/__manifest__.py
brian10048: Resolved with #411

fieldservice_sale_stock/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2019 Brian McMaster
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Field Service - Sale Stock',
+    'version': '12.0.0.1.0',
```
bodedra:
```suggestion
    'version': '12.0.1.0.0',
```

bodedra (approved):
Code review. :+1:

fieldservice/models/res_config_settings.py
```diff
@@ -84,6 +86,14 @@ def _onchange_module_fieldservice_stock(self):
...
             self.group_stock_production_lot = True
             self.group_stock_request_order = True
+            if self.module_fieldservice_sale:
+                self.module_fieldservice_sale_stock = True
+
+    @api.onchange('module_fieldservice_sale')
+    def _onchange_module_fieldservice_sale(self):
+        if self.module_fieldservice_sale and \
+           self.module_fieldservice_stock:
+            self.module_fieldservice_sale_stock = True
```
hparfr: Use auto_install = True and depends in fieldservice_sale_stock/\_\_manifest\_\_.py instead of the _onchange.

hparfr (changes requested):

fieldservice/models/res_config_settings.py
brian10048: Thank you! Updated in latest commit

hparfr (approved):

hparfr:
Thanks @brian10048 
IMO we can merge now

brian10048:
From @max3903 previous comment:
> @brian10048 The Field Service tracking logic seems to only apply to service product. I need it to apply to stockable product as well.
> 
> I created 4 stockable products: 2 tracked by sale, 2 tracked by line and got no FSM order when confirming the sale order.
> 

I thought this was related to not having fieldservice_sale_stock module installed, but I am running into the same issue on runbot.  I remember it was working as intended in previous testing, so I am looking to see what got changed and get this working correctly before merge

brian10048:
@max3903 @hparfr @bodedra Any suggestions that might point me in the right direction to fix this?

I rebased my branch to reflect any recent changes to fieldservice_account module, but this did not fix the problem.

When testing on a local Enterprise install, the pickings from stock products are properly linked to a newly created FSM order, but I am not receiving same result on the runbot Community version.

Both versions work as intended for a product type set to service. 

# Enterprise
## FSM Tracking Set to per Sale Order
image

## Confirmed SO
Notice the linked FSM order smart button
image

## Pickings linked to FSM Order
image



# Community
## FSM Tracking Set to per Sale Order
image

## Confirmed SO
No FSM Order linked
image

fieldservice_sale/readme/CONFIGURE.rst
```diff
@@ -1,12 +1,27 @@
...
+
+The field service tracking of a product defines how it generates a field service
+order if the product is sold via sale order.
+
+To configure a product that generates a unique field service order on each
+sale order line:
 
 * Go to Sales > Catalog > Products
 * Create or select a product
 * Set the Product Type to 'Service' under General Information tab
```
max3903:
```suggestion
```

max3903:
@brian10048 As we use templates on the product, the "Manager Order Templates" option should be activated in Field Service.

max3903:
@brian10048 I still can't get any FSM orders to be created:
http://3400685-403-2ed3cb.runbot1.odoo-community.org/web?#id=23&action=480&model=sale.order&view_type=form&menu_id=329

Product 1 & 2 are set to create 1 FSM order per sale order.
Product 3 & 4 are set to create 1 FSM order per sale order line.

What am I missing?

brian10048:
@max3903. I'm having the same problem on runbot, but not when I'm testing on an Enterprise version.

Can you help me figure out what I'm missing?

max3903:
@brian10048 Sure. I will debug this tomorrow morning.

max3903:
@brian10048 Please review https://github.com/brian10048/field-service/pull/9

I believe the issue is with the `fieldservice_sale_recurring` module. I was not able to reproduce the issue on CE and EE when it is not installed.

brian10048:
> 
> @brian10048 Please review [brian10048#9](https://github.com/brian10048/field-service/pull/9)
> 
> I believe the issue is with the `fieldservice_sale_reccuring` module. I was not able to reproduce the issue on CE and EE when it is not installed.

OK, thank you! That has been driving me crazy. 
Makes sense because I never installed fieldservice_sale_recurring on my EE test.

Should I look into and fix the fieldservice_sale_recurring module as part of this PR or should we get this merged and work on that separately?

max3903:
@brian10048 Let's merge this one first.

I made another pull request: https://github.com/brian10048/field-service/pull/10

max3903 (approved):

brian10048:
Thanks @max3903 

I can work on fixing `fieldservice_sale_recurring` over the next few days as time permits.

max3903:
@brian10048 Here is the migration script: https://github.com/brian10048/field-service/pull/11

PR #404, opened 2019-11-21
[FIX] fieldservice_account: Extend FSM Account Wizard

osi-scampbell:
FSM Account adds customer_id to FSM Location and because it is a mandatory field we are unable to convert a Contact into an FSM Location without extending the wizard. This is the same process we use to account for Inventory Location when it is added to FSM Location in FSM Stock


max3903:
@osi-scampbell Wizards go to the `wizard` directory.

fieldservice_account/models/fsm_wizard.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (C) 2018 - TODAY, Open Source Integrators
```
max3903:
```suggestion
# Copyright (C) 2019 Open Source Integrators
```

bodedra (approved):

max3903 (approved):

PR #406, opened 2019-11-22
[IMP] fieldservice_sale: Autofill the FS location in Sale Orders

clementmbr:
Auto-fill the FS Location in Sale Orders.

Choose the **delivery address** if it is linked to a FS location, else choose the **commercial_partner** of the SO's partner_id if it is linked to a FS location, else do nothing.

I noticed that there is the boolean `fsm_location` (named 'Is a FS location') in `res.partner`, but this field is not consistent. For instance, when you create a new FS Location manually (Master Data > Locations > Create), the related new partner has not `fsm_location` set to `True`.

Moreover, the search in the `fsm.location` table is necessary in order to fill the requested `fsm_location_id`, so this search itself allows to check if the SO's partner is a FS location or not, with no need of the boolean `fsm_location`.


max3903:
> I noticed that there is the boolean `fsm_location` (named 'Is a FS location') in `res.partner`, but this field is not consistent. For instance, when you create a new FS Location manually (Master Data > Locations > Create), the related new partner has not `fsm_location` set to `True`.

@osi-scampbell Can you look into this and fix it?

hparfr (approved):
Problem 1: If the partner is a FSM location, it should be selected directly.
Problem 2: The field is not reset when you switch to another customer with no fsm_location.

max3903:
Problem 3: Why searching twice, when one search can be enough.

clementmbr:
@hparfr and @max3903, thank you very much for your reviews !
I changed the way to make this `onchange` method, which might resolve the 3 issues you have raised.
I also added tests to cover this new method... hope you like it ! :)

fieldservice_sale/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2019 Ecosoft Co., Ltd (http://ecosoft.co.th/)
```
max3903:
```suggestion
# Copyright (C) 2019 Clément Mombereau (Akretion)
```

clementmbr: changed ! Thanks.

max3903:
@clementmbr You should also add yourself to the list of contributors...

hparfr (approved):
Thanks !

brian10048 (approved):

PR #424, opened 2019-12-11
[IMP] fieldservice_sale: Add test

brian10048:
b1a8cefd0670736fe18549b8ff31f279dc9fcbfa fixes the bugs found while validating the tests
3ab457774b0b83ce8871eff3f0529cbd5e951670 adds tests for `fieldservice_sale`


brian10048:
@max3903 I'm running into some trouble with the tests for `fieldservice_sale`

I want to run the tests without/or before `fieldservice_isp_account` is installed in the test environment because it requires setting a timesheet on the FSM order before calling action_confirm for that FSM order.   [See here](https://travis-ci.org/OCA/field-service/jobs/623834867#L2323)

Is there something simple I can do to solve this?

max3903:
> Is there something simple I can do to solve this?

In your test, I would check if `fieldservice_isp_account` is installed and test accordingly.

brian10048:
> 
> 
> > Is there something simple I can do to solve this?
> 
> In your test, I would check if `fieldservice_isp_account` is installed and test accordingly.

OK. That wasn't the simple answer I hoped for! Thank you

brian10048:
#486

Issue #391, opened 2019-11-13
[RFC] fieldservice_account


max3903:
Some organizations allows their FSM persons to have access to the customer invoices.

# fsm.order

- [x] Add a one2many field "Invoices" (invoice_ids) to account.invoice using the field fsm_order_id
- [x] Show a smart button if there is more than one invoice related to the fsm order

brian10048:
In #387 I did an `invoice_id` as you mentioned would be useful to add.  It could be changed to invoice_ids and accomplish the same thing.

Are there any security groups/settings to consider that will allow/restrict certain groups or even user specific access to the invoices?

max3903:
@brian10048 In fieldservice_account, the FSM user needs to have read access to invoice.

In #381, they will accept payments so we will need to give them more permissions in `fieldservice_account_payment`.

dreispt:
The design does not mention where to display the Invoices, I suggested adding it inside a new Notebook Page "Invoices".

murtuzasaleh:
#430

max3903:
@dreispt What about a smart button instead? `fieldservice_account_analytic` is adding an "Accounting" tab and that might be confusing to have Accounting and Invoices...

brian10048:
I like the smart button too.  I would think many times only 1 invoice might be linked so a button would be easier than finding it in another tab

dreispt:
The smart button will go again the requirement in #381 for the Order to have "Register Payment" buttons for the related Invoices. Should we then remove that requirement?

max3903:
> The smart button will go again the requirement in #381 for the Order to have "Register Payment" buttons for the related Invoices. Should we then remove that requirement?

Done

Issue #381, opened 2019-11-04
[RFC] fieldservice_account_payment


max3903:
Depends: fieldservice_account with #391
License: AGPL-3

This module allows the worker to accept payments during his work at the location.

# Usage

* Go to Field Service
* Select or create a Field Service order
* Click on the "Register Payment" button
* Select the Payment Method
* Enter the amount
* Validate

# Technical Specifications

## account.payment

- [x] Add a many2one field "FSM Order" (fsm_order_id) to fsm.order

## fsm.order

- [ ] ~~On the list of invoices (invoice_ids), add a button "Register Payment" to register the payment. Set the payment fields with:~~
* ~~fsm_order_id = id of the order~~
* ~~partner_id = using the customer on the invoice~~
* ~~partner_type = customer~~
* ~~payment_date = today~~
- [x] Add a smart button to display the list of payments related to the order
- [x] Add the required access rights for the worker to create payments

murtuzasaleh:
#423

PR #430, opened 2019-12-12
[IMP][12.0] fieldservice_account

murtuzasaleh:
#391 

Some organizations allows their FSM persons to have access to the customer invoices.

### DIFF
--- a/fieldservice_account/migrations/12.0.3.0.0/post-migration.py
+++ b/fieldservice_account/migrations/12.0.3.0.0/post-migration.py
@@ -1,0 +1,13 @@
+# Copyright (C) 2019, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(env, version):
+    if not version:
+        return
+
+    env.execute("""INSERT INTO fsm_order_account_invoice_rel
+                SET (fsm_order_id, invoice_id) VALUES (
+                SELECT fsm_order_id, id
+                FROM account_invoice
+                WHERE fsm_order_id IS NOT NULL);""")
--- a/fieldservice_account/models/account_invoice.py
+++ b/fieldservice_account/models/account_invoice.py
@@ -8,29 +8,21 @@ from odoo import fields, models, api
 class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
 
-    fsm_order_ids = fields.One2many(
-        comodel_name='fsm.order',
-        string='FSM Orders',
-        compute='_compute_fsm_order_ids',
-        readonly=True, copy=False)
+    fsm_order_ids = fields.Many2many(
+        'fsm.order', 'fsm_order_account_invoice_rel',
+        'invoice_id', 'fsm_order_id', string='FSM Orders')
     fsm_order_count = fields.Integer(
         string='FSM Order Count',
-        compute='_compute_fsm_order_ids', readonly=True)
-    fsm_order_id = fields.Many2one('fsm.order', 'FSM Order')
+        compute='_compute_fsm_order_count', readonly=True)
 
-    @api.depends('invoice_line_ids.fsm_order_id')
-    def _compute_fsm_order_ids(self):
+    @api.depends('fsm_order_ids')
+    def _compute_fsm_order_count(self):
         for invoice in self:
-            orders = invoice.invoice_line_ids.mapped('fsm_order_id')
-            invoice.fsm_order_ids = orders
-            invoice.fsm_order_count = len(orders)
+            invoice.fsm_order_count = len(invoice.fsm_order_ids)
 
     @api.multi
     def action_view_fsm_orders(self):
-        # fetch all orders:
-        # linked to this account
-        line_ids = self.invoice_line_ids.ids
         action = self.env.ref(
             'fieldservice.action_fsm_dash_order').read()[0]
-        action['domain'] = [('invoice_line_id', 'in', line_ids)]
+        action['domain'] = [('invoice_ids', 'in', self.id)]
         return action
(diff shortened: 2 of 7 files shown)


murtuzasaleh:
@max3903 Can you please review this PR.

osi-scampbell (approved):

max3903:
@murtuzasaleh Let's got with a smart button instead of a tab and you are missing the security file to allow read access to invoices for FSM User.

fieldservice_account/views/fsm_order.xml
```diff
@@ -7,6 +7,29 @@
...
+                            <field name="company_id" groups="base.group_multi_company" options="{'no_create': True}"/>
+                            <field name="user_id"/>
+                            <field name="date_due"/>
+                            <field name="origin"/>
+                            <field name="amount_untaxed_invoice_signed" string="Tax Excluded" sum="Total"/>
+                            <field name="amount_tax_signed" string="Tax" sum="Total"/>
+                            <field name="amount_total_signed" string="Total" sum="Total"/>
+                            <field name="residual_signed" string="Amount Due" sum="Amount Due"/>
+                            <field name="currency_id" invisible="1"/>
+                            <field name="state"/>
```
dreispt: This is for information of the FSM Worker, so maybe we don't need all these fields here.
Implementation is correct, but we might want to refine the UX design in the future.

dreispt (approved):

murtuzasaleh:
> @murtuzasaleh Let's got with a smart button instead of a tab and you are missing the security file to allow read access to invoices for FSM User.

The smart button will go again the requirement in #381 for the Order to have "Register Payment" buttons for the related Invoices. Should I then remove that code from the fieldservice_account_payment module?

fieldservice_account/views/fsm_order.xml
dreispt: For clarity, you can  leave it as it is, we can adjust later if needed.

fieldservice_account/models/account_invoice.py
```diff
@@ -16,6 +16,7 @@ class AccountInvoice(models.Model):
     fsm_order_count = fields.Integer(
         string='FSM Order Count',
         compute='_compute_fsm_order_ids', readonly=True)
+    fsm_order_id = fields.Many2one('fsm.order', 'FSM Order')
```
max3903:
```suggestion
```
We already have `fsm_order_ids` above.

max3903:
@wolfhall @hparfr @brian10048 @dreispt I am starting to get confused with the relationship between FSM orders and invoices. Couple commits ago we removed fsm_order_id from invoice and now we are adding it again.

Scenario 1:
We confirm a SO and it generates a delivery order, FSM orders and invoices. The FSM worker can get the payment of the invoice from the FSM order.

--> 1 invoice can be linked to multiple FSM orders.

Scenario 2:
When the FSM order is completed, the user can invoice the timesheets and contractor costs.

--> 1 FSM order can generate one invoice.

If we mix those 2 scenario, we can have 2 invoices for one FSM order (one for the inventory item being delivered, one for the contractor cost/timesheet of the installation).

What do you think of having a many2many relation, instead of 2 one2many ones?

brian10048:
If there is one thing for certain we all have unique workflows for invoicing our field services.  If a many2many allows more flexibility for all scenarios, then that sounds good to me.

fieldservice_account/models/account_invoice.py
max3903: @murtuzasaleh Remove this field and change fsm_order_ids above to a many2many.

fieldservice_account/models/fsm_order.py
```diff
@@ -17,3 +17,7 @@ class FSMOrder(models.Model):
         related="invoice_line_id.invoice_id",
         readonly=True,
     )
+    invoice_ids = fields.One2many('account.invoice',
+                                  'fsm_order_id',
+                                  'Invoices'
+                                  )
```
max3903: @murtuzasaleh many2many as well

fieldservice_account/views/account_invoice.xml
```diff
@@ -14,6 +14,9 @@
                     <field name="fsm_order_count" widget="statinfo" string="FSM Orders"/>
                 </button>
             </div>
+            <field name="reference" position="after">
+                <field name="fsm_order_id"/>
+            </field>
```
max3903: @murtuzasaleh You can remove this. We will use the smart button.

fieldservice_account/views/fsm_order.xml
```diff
@@ -7,6 +7,29 @@
...
+                        </tree>
+                    </field>
+                </page>
+            </notebook>
```
max3903: @murtuzasaleh Let's use a smart button instead. Show it if there is at least one invoice.

max3903:
@murtuzasaleh Please review and merge https://github.com/murtuzasaleh/field-service/pull/4

max3903 (approved):

max3903:
@murtuzasaleh Can you provide the migration script please?

fieldservice_account/migrations/12.0.3.0.0/post-migration.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (C) 2019, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(env, version):
+    if not version:
+        return
+
+    env.execute("""INSERT INTO fsm_order_account_invoice_rel
+                SET (fsm_order_id, account_id) VALUES (
```
max3903:
```suggestion
                SET (fsm_order_id, invoice_id) VALUES (
```

brian10048 (approved):

PR #436, opened 2019-12-14
[IMP] fieldservice_sale_recurring: Add test

brian10048:
Depends on #486


bodedra (approved):
:+1:

dreispt (approved):
Tests should not depend on demo data, they should build their test data.
This allows them to be used on non demo databases, such as a test copy of production.
Still, while not perfect, this is better than no tests at all...

brian10048:
Thanks @dreispt . That is good to know for the future.

marcelsavegnago (approved):

PR #438, opened 2019-12-16
[MIG] Migration Script

osi-scampbell:
This PR adds migration script to fieldservice_analytic_account. We store the analytic_account_id in a temporary column before we migrate. After we migrate, we update analytic_account_id with the value stored in that temporary column.

### DIFF
--- a/fieldservice/i18n/de.po
+++ b/fieldservice/i18n/de.po
@@ -2490,6 +2490,11 @@ msgid "Responsible User"
 msgstr ""
 
 #. module: fieldservice
+#: selection:fsm.stage,stage_type:0
+msgid "Route"
+msgstr ""
+
+#. module: fieldservice
 #: model:ir.model.fields,field_description:fieldservice.field_fsm_location__sale_order_count
 #: model:ir.model.fields,field_description:fieldservice.field_fsm_person__sale_order_count
 msgid "Sale Order Count"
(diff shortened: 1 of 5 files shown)


fieldservice_account_analytic/migrations/12.0.2.1.1/pre-migration.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2019, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    cr.execute("SELECT id,analytic_account_id FROM fsm_location")
```
max3903:
```suggestion
    cr.execute("SELECT id,analytic_account_id FROM fsm_location WHERE analytic_account_id IS NOT NULL")
```

fieldservice_account_analytic/migrations/12.0.2.1.0/pre-migration.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2019, Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    cr.execute("SELECT id,analytic_account_id FROM fsm_location WHERE analytic_account_id IS NOT NULL")
```
max3903:
```suggestion
    cr.execute("""
        SELECT l.id,l.analytic_account_id, r.company_id
        FROM fsm_location AS l
        INNER JOIN res_partner AS r ON r.id = l.partner_id
        WHERE l.analytic_account_id IS NOT NULL""")
```

PR #439, opened 2019-12-17
[IMP][12.0] fieldservice - Timeline 

hpatelserpentcs:
### FieldService: Scheduled Start - End times
Improved the calculation of scheduled dates and duration

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -240,22 +240,81 @@ class FSMOrder(models.Model):
 
     def _calc_scheduled_dates(self, vals):
         """Calculate scheduled dates and duration"""
-        if 'scheduled_date_end' in vals:
-            date_to_with_delta = fields.Datetime.from_string(
-                vals.get('scheduled_date_end')) - \
-                timedelta(hours=self.scheduled_duration)
-            vals['scheduled_date_start'] = str(date_to_with_delta)
-        if 'scheduled_duration' in vals:
-            date_to_with_delta = fields.Datetime.from_string(
-                vals.get('scheduled_date_start', self.scheduled_date_start))\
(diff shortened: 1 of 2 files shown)


wolfhall (approved):
Looks good- ready for functional testing

brian10048:
Can you explain more what we are trying to accomplish with this PR?  Does it help calculate more precise times?

The test is failing [here](https://github.com/OCA/field-service/blob/12.0/fieldservice/tests/test_fsm_order.py#L64).  I'm not sure if the failure is due to this change or if the test case has errors.  I don't understand the reason for Line 61-62 in that test and it may be throwing things off

hparfr (changes requested):
Thanks for this contribution.

Please refactor it: this function is too long and contains code duplication.

hpatelserpentcs:
Hello @brian10048 

> Can you explain more what we are trying to accomplish with this PR? Does it help calculate more precise times?
> 
> The test is failing [here](https://github.com/OCA/field-service/blob/12.0/fieldservice/tests/test_fsm_order.py#L64). I'm not sure if the failure is due to this change or if the test case has errors. I don't understand the reason for Line 61-62 in that test and it may be throwing things off

This calcultion for updates to the fms order form data and also update the timeline. when we updated the duration from timeline it not affected to order.

I will update UT.

hpatelserpentcs:
> Thanks for this contribution.
> 
> Please refactor it: this function is too long and contains code duplication.

I will improve the code

hpatelserpentcs:
@max3903 

I have updated code and UT please review it.

CC @smangukiya @brian10048 @hparfr

smangukiya (approved):
The code looks good to me. Not Tested.

brian10048 (approved):

PR #432, opened 2019-12-13
[12.0] [ADD] fieldservice_route

nikul-serpentcs:
Issue: #380

Issue #390, opened 2019-11-13
[RFC] fieldservice_route_vehicle


max3903:
Depends on fieldservice_vehicle and fieldservice_route

This module allows you to set the vehicle on the routes and day routes.

# Configuration

* Go to Field Service > Master Data > Routes
* Create or select a route
* Set the Vehicle

# Usage

* Go to Field Service > Operations > Day Routes
* Create or select a day route
* Set the vehicle

# Technical Specifications

## fsm.route

- [x] Add a many2one field "Vehicle" (fsm_vehicle_id) to fsm.vehicle

## fsm.route.dayroute

- [x] Add a many2one field "Vehicle" (fsm_vehicle_id) to fsm.vehicle
- [x] Add an onchange on the route field to set the vehicle of the route to the day route:
```
self.vehicle_id = self.route_id.vehicle_id.id
```

brian10048:
LGTM

When can we implement [geospatial 223](https://github.com/OCA/geospatial/pull/223), add a lasso tool for selecting orders on the map, and create an action to optimize the driving using something like [this](https://developers.google.com/optimization/routing/vrp)?  :grin:

max3903:
Aha, I will ask Santa if he is interested and has budget... ;-)

brian10048:
@max3903 if you can ballpark an estimate please send me an email! 

:santa: might be more efficient if he had it!

max3903:
@brian10048 I would need to investigate a bit to provide an estimate.

CC: @wolfhall

wolfhall:
Brian.  This is also on my wishlist :)   
I worked some 18 years with VRP stuff and my team  designed our own algorithms.   I was hoping to plug this in before the end of the year but I don't have funding.

damendieta:
Hi, I'd like to help with this issue, could you please tell me where can I download the base_geolocalize module, it's on the dependencies, but I only found it for v8.

Thanks a lot.

Daniel.

bodedra:
@damendieta It's Odoo standard core module and you may find it under addons folder. Here is ref: https://github.com/odoo/odoo/tree/12.0/addons/base_geolocalize

Issue #380, opened 2019-11-04
[RFC] fieldservice_route


max3903:
Depends: fieldservice
License: AGPL-3

A Field Service route is the sorted list of orders that one worker/vehicle will perform during a day.

This module allows you to manage your Field Service routes: when assigning an order to a worker and scheduling it, a route is automatically created for that worker and day.

# Configuration

* Go to Field Service > Configuration > Stages
* Define the workflow of your routes by creating new stages whose type is "Route"
* Go to Field Service > Configuration > Routes
* Create your routes by setting their name and selecting their territory

# Usage

* Go to Field Service
* Create or select an order
* Assign it to a worker and schedule it
* Go to Field Service > Dashboard > Day Routes. A new record has been created.

# Technical Specifications

## fsm.stage

- [x] Add a new stage_type "Route" (route)

## fsm.route.day

- [x] Add the object
- [x] Add a selection field "Name" with the 7 days of the week
- [x] Add the data with the 7 records

## fsm.route

- [x] Add a char field "Name" (name)
- [x] Add a many2one field "Territory" (territory_id) to fsm.territory
- [x] Add a many2one field "Person" (fsm_person_id) to fsm.person
- [x] Add a many2many field "Days" to fsm.route.day

## fsm.location

- [x] Add a many2one field "Route" (fsm_route_id) to fsm.route

## fsm.route.dayroute

- [x] Move the `fsm.route` object from the fieldservice module to this module and rename `fsm.route` current object to `fsm.route.dayroute`
- [x] Add a many2one field "Route" (route_id) to fsm.route
- [x] Add a many2one field "Team" (team_id) to fsm.team. Only shown if the user is in the FSM Team group.
- [x] Add a many2one field "Stage" (stage_id) to fsm.stage filtered with stage_type = route
- [x] Add a many2one field "Territory" (territory_id)
- [x] Add 2 float fields "Longitude" (longitude) and "Latitude" (latitude)
- [x] Add a many2one field "Last Location" (last_location_id) to fsm.location
- [x] Add a datetime field "Planned Start Time" (date_start_planned). The time of day when the worker starts to work, it needs to consider time zone and schedule for this worker (use default worker setting).
- [x] Add a many2one field "Start Location" (start_location_id) to fsm.location. The worker's day starts at this location. It can be a home, office, etc. (optional -use default worker setting)
[...]

PR #440, opened 2019-12-17
[12.0][IMP] fieldservice: Remove route

murtuzasaleh:
Removed fsm.route object from here.

Related to #432


murtuzasaleh:
@max3903 @dreispt Can you please review this PR.

dreispt:
If this is removed, will it be replaced somewhere?
There should be an explanation of the reasons to remove this, for future reference.

murtuzasaleh:
> If this is removed, will it be replaced somewhere?
> There should be an explanation of the reasons to r

fsm.route will be added in fieldservice_route module #380
Cc @max3903

murtuzasaleh:
@max3903 fieldservice_account_analytic and fieldservice_vehicle module have fsm.route model reference.
https://github.com/OCA/field-service/blob/12.0/fieldservice_vehicle/models/fsm_route.py
https://github.com/OCA/field-service/blob/12.0/fieldservice_vehicle/views/fsm_route.xml
https://github.com/OCA/field-service/blob/12.0/fieldservice_account_analytic/views/fsm_route.xml

Do we have to add a fieldservice_route module in the dependency?

dreispt:
IMO we should do all the refactoring in the same PR (removing old model, adding new model, changing referencing Models).

I'm aware these modules are in Beta, but is it OK for everyone involved to have this breaking change?

brian10048:
There is almost no route functionality in base field service as it is, so it is fine by me to remove. 

I also think it would make since as one PR with a handful of commits for each major step in the refactoring/addition of routes

murtuzasaleh:
@max3903 @dreispt Can we have to change or update this action? Because it blocks the base flow.
https://github.com/OCA/field-service/blob/12.0/fieldservice/views/fsm_schedule.xml#L28

dreispt:
After Brian’s comment, I’m OK with this PR.
Please get a green build and you’ll have my OK.

brian10048:
> 
> 
> @max3903 fieldservice_account_analytic and fieldservice_vehicle module have fsm.route model reference.
> https://github.com/OCA/field-service/blob/12.0/fieldservice_vehicle/models/fsm_route.py
> https://github.com/OCA/field-service/blob/12.0/fieldservice_vehicle/views/fsm_route.xml
> https://github.com/OCA/field-service/blob/12.0/fieldservice_account_analytic/views/fsm_route.xml
> 
> Do we have to add a fieldservice_route module in the dependency?

It looks like you guys are making fieldservice_vehicle_route, so the vehicle related references to routes can go there.

To me it doesn't seem necessary enough to add a dependency to fieldservice_account_analytic just to show the billing_id field in the tree view.

max3903:
> @max3903 fieldservice_account_analytic and fieldservice_vehicle module have fsm.route model reference.
> https://github.com/OCA/field-service/blob/12.0/fieldservice_vehicle/models/fsm_route.py
> https://github.com/OCA/field-service/blob/12.0/fieldservice_vehicle/views/fsm_route.xml
> https://github.com/OCA/field-service/blob/12.0/fieldservice_account_analytic/views/fsm_route.xml
> 
> Do we have to add a fieldservice_route module in the dependency?

They should not depend on `fieldservice_route`. See #390

max3903 (approved):

brian10048 (approved):

dreispt (approved):

murtuzasaleh:
> @max3903 @dreispt Can we have to change or update this action? Because it blocks the base flow.
> https://github.com/OCA/field-service/blob/12.0/fieldservice/views/fsm_schedule.xml#L28

@max3903 @dreispt @brian10048  We need these changes functionally?

dreispt:
@murtuzasaleh I just merged this PR. That change, if OK, should go in another PR, sorry.

murtuzasaleh:
> @murtuzasaleh I just merged this PR. That change, if OK, should go in another PR, sorry.

ok

PR #481, opened 2020-01-17
[IMP] fieldservice: Fix team stages

brian10048:
Only show stages for the appropriate team on order kanban and form views

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2018 - TODAY, Open Source Integrators
+# Copyright (C) 2018 Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from datetime import datetime, timedelta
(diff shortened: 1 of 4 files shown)


bodedra (approved):
LGTM :+1:

fieldservice/models/fsm_order.py
```diff
@@ -183,10 +183,12 @@ def _onchange_location_id_customer(self):
...
-        stage_ids = self.env['fsm.stage'].\
-            search([('stage_type', '=', 'order'),
-                    ('company_id', '=', self.env.user.company_id.id)])
-        return stage_ids
+        search_domain = [('stage_type', '=', 'order')]
+        if self.env.context.get('default_team_id'):
+            search_domain = [
+                '&', ('team_ids', 'in', self.env.context['default_team_id'])
+            ] + search_domain
+        return stages.search(search_domain, order=order)
```
hparfr, line 191: Can we have stages with no team_id ?

fieldservice/views/fsm_order.xml
```diff
@@ -21,7 +21,9 @@
                             attrs="{'invisible': [('stage_id', 'in', (%(fieldservice.fsm_stage_completed)d, %(fieldservice.fsm_stage_cancelled)d))]}"/>
                     <field name="stage_id" widget="statusbar"
                            options="{'fold_field': 'fold'}"
-                           domain="[('stage_type', '=', 'order')]"/>
+                           domain="['&amp;', '|',
```
hparfr: You can shorten it to : (stage_type', '=', 'order'), ('team_ids', 'in', (team_id', False))

hparfr (changes requested):

fieldservice/models/fsm_order.py
brian10048: If you aren't using teams in fieldservice, everything will be set to the default team and team fields are hidden.  So all stages would have a team.  team_ids is not a required field on stage, so maybe it should be? It's a required field on the order

hparfr: If stage is not required to have a team, then it can be good to filter team_ids = false.

About adding a required, I don't know, usually I prefer to not having required field if you don't need it absolutely (it can be a burden during tests).

hparfr (approved):

PR #482, opened 2020-01-17
[FIX] fieldservice_route: Issue with locale

max3903:


max3903:
Ping

brian10048 (approved):
Code review, not tested

murtuzasaleh:
Code Review LGTM but when I create Day Routes on Runbot it will give me an error because of the `fieldservice_route_vehicle` module `_get_default_vehicle` method.

Below is the SS of it.

Day_route

max3903:
@murtuzasaleh Should be fixed in #496

fieldservice_route/models/fsm_route.py
```diff
@@ -13,3 +12,19 @@ class FSMRoute(models.Model):
...
+        :return: True if the route runs on the date, False otherwise.
+        """
+        if date:
+            day_index = date.weekday()
+            day = self.env.ref(
+                'fieldservice_route.fsm_route_day_' + str(day_index))
+            if day in self.day_ids:
+                return True
+            else:
+                return False
```
dreispt:
```suggestion
            return bool(day in self.day_ids)
```

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -13,6 +14,13 @@ class FSMRouteDayRoute(models.Model):
     def _get_default_person(self):
         return self.route_id.fsm_person_id.id or False
 
+    @api.model
+    def _get_default_date_start_planned(self):
```
dreispt, line 18: s/date/datetime ?

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -47,6 +58,9 @@ def _compute_order_count(self):
                                 string='Orders')
     order_count = fields.Integer(string='Number of Orders',
                                  compute=_compute_order_count)
+    order_max = fields.Integer(
+        related='route_id.max_order', string="Capacity",
```
dreispt, line 62: It would be best to be consistent and use the same filed name as in the original related model.

dreispt (approved):
LGTM, minor comments/suggestions

nelsonramirezs (approved):

fieldservice_route/models/fsm_route_dayroute.py
max3903: @dreispt The field is named `date_start_planned`.

nelsonramirezs (approved):

PR #491, opened 2020-01-29
[IMP] fieldservice: Minor changes

osi-scampbell:
When a user changes the Internal Reference on a FSM Location, the complete_name field is not updated. Since it is part of the complete_name field, I believe it should update when changed.

Ticket #1499

When a user changes the Template we only set Type if it is False. We also want to append to the existing description instead of overwriting it when copyting notes.

Ticket #1679

When a user pulls up a tree view, enables a group_by filter, selects a record, then selects a smartbutton, if that group_by field is not on the corresponding smart_button object then a KeyError will be thrown. This is resolved by removing the group_by key from the context dictionary after it is copied.

Ticket #1336

On FSO's when a user sets the Location, the directions are then copied into the Location Directions on the FSO properly. However, whenever the call_notes() is called, it duplicates the Location Directions. The only thing that should be in this field are the directions from the assigned FSM Location.

Ticket #1854

On FSO's users were moving the Kanban cards and not triggering the logic that is stored in the buttons that move the stage on the form view. We prevent this by adding a boolean field that is returned True only when the button is clicked. If it is not through the button then we will throw a UserError and move the Kanban card back to its original stage

Ticket #1481

### DIFF
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -61,7 +61,7 @@ class FSMLocation(models.Model):
                                group_expand='_read_group_stage_ids',
                                default=lambda self: self._default_stage_id())
 
-    @api.depends('partner_id.name', 'fsm_parent_id.complete_name')
+    @api.depends('partner_id.name', 'fsm_parent_id.complete_name', 'ref')
     def _compute_complete_name(self):
         for loc in self:
             if loc.fsm_parent_id:
@@ -253,21 +253,21 @@ class FSMLocation(models.Model):
         either be a in a list or in a form view, if there is only one
         contact to show.
         '''
-        for location in self:
+        for contact in self:
             action = self.env.ref('contacts.action_contacts').\
                 read()[0]
-            contacts = self.get_action_views(1, 0, location)
+            contacts = self.get_action_views(1, 0, contact)
             action['context'] = self.env.context.copy()
+            action['context'].update({'group_by': ''})
             action['context'].update({'default_service_location_id': self.id})
-            if len(contacts) == 1:
-                action['views'] = [(self.env.ref('base.view_partner_form').id,
+            if len(contacts) == 0 or len(contacts) > 1:
+                action['domain'] = [('id', 'in', contacts.ids)]
+            elif contacts:
+                action['views'] = [(self.env.
+                                    ref('base.' +
+                                        'view_partner_form').id,
                                     'form')]
                 action['res_id'] = contacts.id
-                action['context'].update({'active_id': contacts.id})
-            else:
-                action['domain'] = [('id', 'in', contacts.ids)]
-                action['context'].update({'active_ids': contacts.ids})
-                action['context'].update({'active_id': ''})
             return action
 
     @api.multi
@@ -288,6 +288,7 @@ class FSMLocation(models.Model):
                 read()[0]
             equipment = self.get_action_views(0, 1, location)
             action['context'] = self.env.context.copy()
+            action['context'].update({'group_by': ''})
             action['context'].update({'default_location_id': self.id})
             if len(equipment) == 0 or len(equipment) > 1:
                 action['domain'] = [('id', 'in', equipment.ids)]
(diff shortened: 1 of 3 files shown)


fieldservice/models/fsm_location.py
```diff
@@ -256,18 +256,18 @@ def action_view_contacts(self):
...
+                                        'view_partner_form').id,
                                     'form')]
-                action['res_id'] = contacts.id
-                action['context'].update({'active_id': contacts.id})
-            else:
-                action['domain'] = [('id', 'in', contacts.ids)]
-                action['context'].update({'active_ids': contacts.ids})
-                action['context'].update({'active_id': ''})
+                action['res_id'] = locations.id
             return action
```
max3903, line 271: @osi-scampbell I am not sure to follow you here: This is the method to view contacts from the location form view and based on the number of contacts, we display the form or list view of contacts... The variable name seems correct to me.

osi-scampbell: @max3903 You are correct this variable is named improperly

max3903 (approved):

fieldservice/models/fsm_order.py
```diff
@@ -324,17 +334,18 @@ def copy_notes(self):
                         self.description = (self.equipment_id.notes + '\n ')
         if self.location_id:
             s = self.location_id.direction
+            self.location_directions
```
brian10048: ?

brian10048: = ?

fieldservice/models/fsm_order.py
```diff
@@ -324,17 +334,18 @@ def copy_notes(self):
...
-                    self.location_directions = (self.location_directions +
-                                                '\n' + s + '\n')
-                else:
-                    self.location_directions = (s + '\n ')
+                self.location_directions = (s + '\n ')
         if self.template_id:
-            self.todo = self.template_id.instructions
+            self.todo += self.template_id.instructions
+        if self.description:
+            self.description = '\n' + old_desc
```
brian10048:
```suggestion
            self.description += '\n' + old_desc
```
?

fieldservice/models/fsm_order.py
```diff
@@ -355,7 +366,8 @@ def _onchange_template_id(self):
             self.category_ids = self.template_id.category_ids
             self.scheduled_duration = self.template_id.hours
             self.copy_notes()
-            self.type = self.template_id.type_id
+            if not self.type:
```
brian10048: Why prevent changing the type if it was already set?
For me, if the template has a type it should be set on the order when the template is changed. New template means new type.  Unless the template has no type and the order already has a type, then we could keep the order type as set

brian10048 (changes requested):

fieldservice/models/fsm_location.py
```diff
@@ -253,21 +253,21 @@ def action_view_contacts(self):
         either be a in a list or in a form view, if there is only one
         contact to show.
         '''
-        for location in self:
+        for contact in self:
```
max3903: @osi-scampbell `self` is a recordset of locations not contacts...

fieldservice/models/fsm_order.py
```diff
@@ -306,37 +314,41 @@ def onchange_scheduled_duration(self):
...
-                if self.location_directions is not False:
-                    self.location_directions = (self.location_directions +
-                                                '\n' + s + '\n')
-                else:
-                    self.location_directions = (s + '\n ')
+                self.location_directions = (s + '\n ')
         if self.template_id:
-            self.todo = self.template_id.instructions
+            if self.template_id.instructions:
+                self.todo += self.template_id.instructions
```
brian10048: This is causing a duplicate of the same instructions or appends a new set of instructions leaving the previous template instructions, which seems incorrect to me.  With a new template comes a new set of instructions

And in the case of nothing in self.todo it raises error when trying to set a template on the form

 ` File "/home/odoo/build/OCA/field-service/fieldservice/models/fsm_order.py", line 346, in copy_notes
    self.todo += self.template_id.instructions
TypeError: unsupported operand type(s) for +=: 'bool' and 'str'`

brian10048 (changes requested):

patrickrwilson (approved):
Looks good.

max3903:
@brian10048 Can we merge this one?

brian10048 (approved):

PR #499, opened 2020-02-05
[FIX] fieldservice_location_builder: Timezone Error + Dependency Error

osi-scampbell:
The 'tz' field on an fsm_location is a selection field. When building a hierarchy on a record with a tz an error is thrown the is fixed with this PR. Another error discovered was customer_id has been moved to fieldservice_account_analytic. The 2 solutions are to have this module depend on fieldservice_account_analytic, or split the customer_id field into a separate module called fieldservice_account_analytic_location_builder similar to how we made fieldservice_stock_location_builder.

UPDATE 02/28: I split the vals on the wizard into a separate method. This makes it easier for us to inherit this method and extend its functionality

### DIFF
--- a/fieldservice_location_builder/__manifest__.py
+++ b/fieldservice_location_builder/__manifest__.py
@@ -9,7 +9,7 @@
     'author': 'Open Source Integrators, Odoo Community Association (OCA)',
     'website': 'https://github.com/OCA/field-service',
     'depends': [
-        'fieldservice'
+        'fieldservice_account_analytic'
     ],
     'data': [
         'wizard/fsm_location_builder_wizard.xml',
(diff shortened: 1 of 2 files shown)


mgosai (approved):

fieldservice_location_builder/__manifest__.py
```diff
@@ -9,7 +9,8 @@
     'author': 'Open Source Integrators, Odoo Community Association (OCA)',
     'website': 'https://github.com/OCA/field-service',
     'depends': [
-        'fieldservice'
+        'fieldservice',
+        'fieldservice_account_analytic'
```
max3903, line 12: @osi-scampbell Why do you need this dependency?

max3903: @osi-scampbell The module that adds the customer_id field should provide the default value so you don't have to depend on it here.

max3903 (approved):

PR #507, opened 2020-02-19
[ADD] fieldservice_size

### LINKED ISSUE #384: [RFC] Order, Location Sizes
In my field service operation it is necessary to track the size of a field service order.  For example, the square foot area of a lawn.

The size of a field service order is typically based on the size of the location being serviced.

My company provides multiple sectors of services to one location which can be grouped using categories.  As a result, a location needs to have a size for each category.
  
Sizes need configurable units of measure.

A field service order template uses the size to calculate the expected duration for a field service order. (Example: each unit of size requires 'x' minutes to complete)

brian10048:
#384 

New module to manage sizes of locations for different orders types

@wolfhall @max3903 Please let me know your thoughts or suggestions for improvement


fieldservice_size/models/fsm_size.py
```diff
@@ -0,0 +1,30 @@
...
+        ('name_uniq', 'unique (name)', "Size name already exists!"),
+    ]
+
+    @api.constrains('is_order_size', 'type_id')
+    def _one_size_per_type(self):
+        size = self.search([
+            ('type_id', '=', self.type_id.id),
+            ('is_order_size', '=', True)
+        ])
+        if len(size) == 2:
```
max3903:
```suggestion
        if len(size) >= 2:
```

max3903 (approved):
LGTM. Not tested.

fieldservice_size/models/fsm_size.py
```diff
@@ -0,0 +1,30 @@
...
+        string='Is the Order Size?',
+        help='The default size for orders of this type')
+
+    _sql_constraints = [
+        ('name_uniq', 'unique (name)', "Size name already exists!"),
+    ]
+
+    @api.constrains('is_order_size', 'type_id')
+    def _one_size_per_type(self):
+        size = self.search([
```
hparfr: Not a requested change, but you may use search_count to get better perfs

fieldservice_size/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+The module allows you to establish sizes for your field service orders.
+Multiple sizes can be configured and established for a location. Order
+size will be updated based on its location and order type.
```
hparfr, line 13: When reading this I don't get what can be  a "size" of FSM Order.
Can you add more explanation about what is it / when we need it ?

hparfr (changes requested):
Please add some explanation on the readme :)

fieldservice_size/models/fsm_size.py
brian10048: Great. Thank you!

fieldservice_size/readme/DESCRIPTION.rst
brian10048: @hparfr I updated the readme to help make the use of sizes more clear.

For me, my specific use is as follows:
We apply fertilizer to a lawn.  The lawn size is measured in square feet units.  Knowing the size of the lawn helps to estimate how many orders to plan for a worker in a day and how many units of fertilizer need to be used at each location.

Thank you for the review and let me know if I can further update the readme

hparfr (approved):

PR #515, opened 2020-02-27
[FIX] fieldservice: False team from template

brian10048:
Fixes bug introduced with #441 

If teams are not enabled, the template's team will be False.  When a user sets a template on the order, the order team is set to False when team is a required field

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -358,4 +358,5 @@ class FSMOrder(models.Model):
             self.scheduled_duration = self.template_id.hours
             self.copy_notes()
             self.type = self.template_id.type_id
-            self.team_id = self.template_id.team_id
+            if self.template_id.team_id:
+                self.team_id = self.template_id.team_id


fieldservice/models/fsm_order.py
```diff
@@ -358,4 +358,5 @@ def _onchange_template_id(self):
             self.scheduled_duration = self.template_id.hours
             self.copy_notes()
             self.type = self.template_id.type_id
-            self.team_id = self.template_id.team_id
+            if self.template_id.team_id:
+                self.team_id = self.template_id.team_id
```
max3903, line 362:
```suggestion
            self.team_id = self.template_id.team_id or False
```

brian10048: @max3903 I don't think this suggestion will work since team_id is a required field of fsm.order?  
The team_id of an order is given a default value before this method and the error happens when this method sets team_id from its default value to False (because there is no team_id for the template).  As a result users cannot save an order with the team_id set to False

I thought about assigning a default value to team_id on the fsm.template and making it a required field so there is always a value from the template, but thought this might be simpler

max3903: @brian10048 Got it.

max3903 (approved):

bodedra (approved):

hparfr (approved):

PR #517, opened 2020-02-28
[ADD] FSM Analytic Location Builder

osi-scampbell:
This PR adds fieldservice_analytic_location_builder. It extends the fieldservice_location_builder module in order to set all child locations Analytic Account

Depends on: #499


patrickrwilson (approved):
Worked Great, LGTM

PR #522, opened 2020-03-13
[FIX] fieldservice: Keep type when changing template

max3903:


murtuzasaleh:
LGTM

murtuzasaleh (approved):

fieldservice/models/fsm_order.py
```diff
@@ -367,6 +367,7 @@ def _onchange_template_id(self):
             self.category_ids = self.template_id.category_ids
             self.scheduled_duration = self.template_id.hours
             self.copy_notes()
-            self.type = self.template_id.type_id
+            if not self.type and self.template_id.type_id:
```
brian10048:
```suggestion
            if self.template_id.type_id:
```

To me the template should set the type unless the template doesn't have a type

max3903: @patrickrwilson What do you think?

patrickrwilson: @max3903 Yes, I agree, if the type field is empty on the fso, and the template has a type, then the type should populate from the template.

patrickrwilson (approved):
LGTM

Issue #524, opened 2020-03-18
[RFC] Customer on Location/Order


max3903:
Sorry to bring the subject back again but I have a requirement to have the customer information on the location and order.

My scenario is that the FSM order is created from the Sale Order validation and the worker needs to have the customer info in addition to the location.

What do you think of:

- [ ] extracting the customer_id field from `fieldservice_account_analytic` but leaving the invoice generation logic
- [ ]  adding it back to `fieldservice_account`

CC: @brian10048 @hparfr @osi-scampbell @wolfhall @rto-odoo

brian10048:
This is not a problem for me.

@hparfr was more concerned about the placement of the customer_id field from what I remember

hparfr:
1)Yes, it was very confusing for multiple customers. Location is enough for most of the cases. 
2) Plus I asked to remove billing id because there was some logic about it. It was not the responsibility of the FSM Order to manage the billing.
3) You can have multiple customers with the same location (the owner / the renter). So binding locations to customers can be wrong.


I can understand the worker needs the name of the client on the Order, but why in the location too?

max3903:
@hparfr 

1) customer_id is not required on FSM order, so you should be fine here.
2) The logic to invoice from a FSM order is not in `fieldservice_account`, so we're good.
3) The location has multiple fields to store the owner, the customer, the contact.

> I can understand the worker needs the name of the client on the Order, but why in the location too?

Because a FSM order can be created manually and the worker needs to know the customer. 

What do you think of renaming the field "Customer" on the location to "Default Customer"? Would that work for your scenario?

hparfr:
Ok, then I will hide all theses fields in a new (custom?) module.

Usually, the base module should be generic and other modules adds features, not remove them.

hparfr:
You don't use field_account_analyitic for this projetct ?

max3903:
@hparfr 

No, this is a delivery service where customer calls, requests X products. Products gets invoiced before delivery. When the products is delivered to the location (address), customer pays.

The location name is the address in the form of "1, Main Drive". No customer name...

What do you think of adding a checkbox in the settings to show the customer_id field using a specific group? Field can be hidden by default until you check the box in the settings. Would that work for you?

osi-scampbell:
[...]

PR #525, opened 2020-03-18
[FIX] fieldservice_account: Issue #524

max3903:
#524


fieldservice_account/models/fsm_location.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = 'fsm.location'
+
+    @api.model
+    def get_default_customer(self):
+        if self.fsm_parent_id:
+            return self.fsm_parent_id.customer_id.id
+        return self.owner_id.id
+
+    customer_id = fields.Many2one(
+        'res.partner', string='Billed Customer', required=True,
```
dreispt: Why does the Customer have to be a required field?

max3903: @dreispt Because the worker/system needs to know who is financially impacted by the service to contact him, to bill him, to get paid, etc...

CC: @wolfhall @patrickrwilson

PR #530, opened 2020-03-19
[IMP] fieldservice: Location Search Full Name + Parent Loc Dir + Stage Notifications

osi-scampbell:
Two relatively small changes that I believe will have a larger impact. For users who have many MDU's with multiple sublocations they might have several locations that look like this:

Location 1 / Building 2 / Unit 100
Location 1 / Building 2 / Unit 101

Location 2 / Building 1 / Unit 100
Location 2 / Building 1 / Unit 101

Currently users view locations in this format so when searching the first Location above, users would enter Loc... and only see Location 1 and Location 2. By using complete_name instead of name they can search by parent locations to find their desired location.

The next change is to Location Directions. Currently if you selected "Location 1 / Building 2 / Unit 100" you would only get the directions on Unit 100. However, if there are notes from Building 2 (ex: 'first 3 floors are being fumigated') and Location 2 (ex: 'Gate code: XXXX') would not be copied onto the FSO. I believe users should be able to see all the notes associated with a given location and then remove what they don't want instead of having to hunt and pick the details needed for the order.


fieldservice/models/fsm_order.py
```diff
@@ -371,3 +368,20 @@ def _onchange_template_id(self):
...
                 self.team_id = self.template_id.team_id
+
+    def _get_location_directions(self, location_id):
+        self.location_directions = ""
+        s = self.location_id.direction
+        parent_location = self.location_id.fsm_parent_id
+        # ps => Parent Location Directions
+        # s => String to Return
+        while parent_location.id is not False:
+            import pdb; pdb.set_trace()
```
max3903:
```suggestion
```

fieldservice/models/fsm_order.py
```diff
@@ -371,3 +368,20 @@ def _onchange_template_id(self):
...
+            ps = parent_location.direction
+            if ps:
+                if s:
+                    s = s + parent_location.direction
+                else:
+                    s = parent_location.direction
```
max3903: @osi-scampbell What about?
```suggestion
                s += parent_location.direction
```

osi-scampbell: Need the if/else to check if s is False or else it throws a TypeMismatch error for 'NoneType' and 'str'

max3903: @osi-scampbell You can initialize `s` with:

```
s = self.location_id.direction or ""
```

fieldservice/models/fsm_order.py
```diff
@@ -371,3 +368,19 @@ def _onchange_template_id(self):
                 self.type = self.template_id.type_id
             if self.template_id.team_id:
                 self.team_id = self.template_id.team_id
+
+    def _get_location_directions(self, location_id):
+        self.location_directions = ""
+        s = self.location_id.direction
```
max3903:
```suggestion
        s = self.location_id.direction or ""
```

max3903 (approved):

patrickrwilson:
@osi-scampbell I'm not completely sold on the default searching by complete name because for us, we have a lot of units for our locations. Some locations can have over a 1000 units. Right now, when we search for the name of the property it only gives us several results, which are easy to read because it only shows a few but by changing the search to look at the complete name, it could result in thousands of sub locations included. 

I feel a better solution might be to add the complete_name to the search filter so a user has the option to use that if they need to search by the full name instead of the name of the location.

What do you think?

osi-scampbell:
@patrickrwilson @max3903 PR updated with an options on Res Config Settings for users to decide if they want to search by the Hierarchy name or not.

patrickrwilson:
@osi-scampbell I like having the option. Maybe i'm missing something but during my functional testing, i've enabled the search by hierarchy option in settings but it doesn't seem to be working for me...
image

image

image

It should show all the locations with 'Test' in the complete name but it isn't.

osi-scampbell:
@patrickrwilson For this PR, we aren't modifying the search view, we are modifying the search that happens when you type letters into a Many2one field for fsm_locations. Can you try search there with and without the option enabled and tell me your results?

patrickrwilson:
> For this PR, we aren't modifying the search view, we are modifying the search that happens when you type letters into a Many2one field for fsm_locations. Can you try search there with and without the option enabled and tell me your results?

@osi-scampbell I see, this works as intended then.

patrickrwilson (approved):
LGTM

brian10048 (approved):

fieldservice/models/res_company.py
```diff
@@ -11,3 +11,6 @@ class ResCompany(models.Model):
         string='Auto-populate Workers on Location based on Territory')
     auto_populate_equipments_on_order = fields.Boolean(
         string='Auto-populate Equipments on Order based on Location')
+    seach_on_complete_name = fields.Boolean(
```
max3903:
```suggestion
    search_on_complete_name = fields.Boolean(
```

fieldservice/models/res_config_settings.py
```diff
@@ -81,6 +81,10 @@ class ResConfigSettings(models.TransientModel):
         string='Auto-populate equipments on Order based on the Location',
         related='company_id.auto_populate_equipments_on_order',
         readonly=False)
+    seach_on_complete_name = fields.Boolean(
```
max3903:
```suggestion
    search_on_complete_name = fields.Boolean(
```

fieldservice/models/res_config_settings.py
```diff
@@ -81,6 +81,10 @@ class ResConfigSettings(models.TransientModel):
...
+        string='Search Location By Hierarchy',
+        related='company_id.seach_on_complete_name',
```
max3903:
```suggestion
        related='company_id.search_on_complete_name',
```

fieldservice/views/res_config_settings.xml
```diff
@@ -49,6 +49,18 @@
                                 </div>
                             </div>
                         </div>
+                        <div class="col-xs-12 col-md-6 o_setting_box">
+                            <div class="o_setting_left_pane">
+                                <field name="seach_on_complete_name"/>
+                            </div>
+                            <div class="o_setting_right_pane">
+                                <label for="seach_on_complete_name"
```
max3903:
```suggestion
                                <field name="search_on_complete_name"/>
                            </div>
                            <div class="o_setting_right_pane">
                                <label for="search_on_complete_name"
```

PR #542, opened 2020-04-02
[IMP] fieldservice_isp_account: Default Set Project and Task

osi-scampbell:
When timesheets are entered they do not show up in the timesheets app because the time entered is not associated with a task or project. This resolves the issue however the user may not know they need to set a Project and Task. I made this module dependent on fsm_project which I believe is necessary however I did not make project_id and project_task_id required. 

Ticket #1234 Low: Timesheet (FSO Entries) – Employee Timesheet Entries on FSO’s don’t show in Timesheet App

### DIFF
--- a/fieldservice_isp_account/__manifest__.py
+++ b/fieldservice_isp_account/__manifest__.py
@@ -11,6 +11,7 @@
     'website': 'https://github.com/OCA/field-service',
     'depends': [
         'fieldservice_account_analytic',
+        'fieldservice_project',
         'hr_timesheet',
     ],
     'data': [
--- a/fieldservice_isp_account/views/fsm_order.xml
+++ b/fieldservice_isp_account/views/fsm_order.xml
@@ -50,7 +50,7 @@
                     <field name="contractor_total" readonly="1"/>
                 </group>
                 <group string="Employee Timesheets">
-                    <field name="employee_timesheet_ids" nolabel="1">
+                    <field name="employee_timesheet_ids" nolabel="1" context="{'default_project_id': project_id, 'default_task_id': project_task_id}">
                         <tree editable="bottom"
                                 string="Timesheet Activities"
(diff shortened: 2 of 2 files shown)


bodedra (approved):

patrickrwilson (changes requested):
This is good but we need a bit more. It would be good to mimic how helpdesk tickets work with their timesheets. Adding a project to the helpdesk team, gives the ticket a default project when creating the ticket.
The same should be for FSM Orders. Add a project field to the FSM Team, and when the dispatch team is selected on the FSM Order, it auto populates the project using the project that is set on the dispatch team.

osi-scampbell:
@patrickrwilson https://github.com/OCA/field-service/pull/543

patrickrwilson (changes requested):
In the employee_timesheet_ids, it seems you need to add the project_id and project_task_id to the view, even if its invisible in order for it to default to what's on the fso.
Add below field name="account_id" invisible="1"
field name="project_id" invisible="1"
field name="project_task_id" invisible="1"

osi-scampbell:
@patrickrwilson You are correct, however the project_id should stay in fieldservice_project and we can't make fieldservice_project depends on fieldservice_isp_account either. So I think this needs to go into a separate method called fieldservice_isp_account_project. Or we leave this as a customization in pavlov repo. @max3903

osi-scampbell:
> @patrickrwilson You are correct, however the project_id should stay in fieldservice_project and we can't make fieldservice_project depends on fieldservice_isp_account either. So I think this needs to go into a separate method called fieldservice_isp_account_project. Or we leave this as a customization in pavlov repo. @max3903

Hold on, I think what I just pushed should work

patrickrwilson (approved):
Yes, this works. LGTM

max3903 (approved):

PR #543, opened 2020-04-02
[IMP] fieldservice_project: Default Project From Team on Order

osi-scampbell:
Paired with #542, automatically default the project on an FSO to whatever is assigned to that team. Similar to the way projects are set on Helpdesk Tickets via Helpdesk Teams.

### DIFF
--- a/fieldservice_project/__manifest__.py
+++ b/fieldservice_project/__manifest__.py
@@ -19,6 +19,7 @@
         'views/fsm_location_views.xml',
         'views/fsm_order_views.xml',
         'security/ir.model.access.csv',
+        'views/fsm_team.xml'
     ],
     'installable': True,
 }
--- a/fieldservice_project/models/__init__.py
+++ b/fieldservice_project/models/__init__.py
@@ -4,3 +4,4 @@ from . import fsm_location
 from . import fsm_order
 from . import project
 from . import project_task
+from . import fsm_team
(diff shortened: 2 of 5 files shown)


patrickrwilson (approved):
Tested and works great

max3903 (approved):

osi-scampbell:
@max3903 @patrickrwilson @brian10048 This isn't the focus of this PR necessarily, but I found this while working on #542. Timesheet entries (account.analytic.line) task_id is filtered on the project_id. Should we do the same here? Or is there a case where an FSO could have a task_id from a different project than the one stored on the same FSO? Could we have task_id on an FSO without a project_id on the FSO? I bring this up because if the task_id is not from the project_id on the FSO then the Timesheet entry will clear the task_id once it is copied over through the context.

patrickrwilson:
> @max3903 @patrickrwilson @brian10048 This isn't the focus of this PR necessarily, but I found this while working on #542. Timesheet entries (account.analytic.line) task_id is filtered on the project_id. Should we do the same here? Or is there a case where an FSO could have a task_id from a different project than the one stored on the same FSO? Could we have task_id on an FSO without a project_id on the FSO? I bring this up because if the task_id is not from the project_id on the FSO then the Timesheet entry will clear the task_id once it is copied over through the context.

@osi-scampbell There shouldn't be a scenario where you have a task that isn't part of the selected project so i think your point is valid.

brian10048 (approved):
No test, code LGTM

PR #549, opened 2020-04-15
[IMP] fieldservice_stock: Request Workflow Adjustment

osi-scampbell:
The following PR makes changes to the fieldservice_stock module to fix a bug that was taking SR's from new SRO's and appending them to existing ones. While this is somewhat the desired behavior, we do not want it automated, and should instead output an error message letting the user know that an existing SRO with matching FSO and Warehouse and is in draft stage exists, and the user should make their SR there. I added a few things to make sure to make sure we do not interfere with the default SR Workflow when it is not associated with an FSO.

Edit: In order to maintain consistency with PR: OCA/stock-logistics-warehouse#874 I have changed the header for the onchange methods in fieldservice_stock

### DIFF
--- a/fieldservice_stock/models/stock_request_order.py
+++ b/fieldservice_stock/models/stock_request_order.py
@@ -11,7 +11,7 @@ class StockRequestOrder(models.Model):
         'fsm.order', string="FSM Order", ondelete='cascade',
         index=True, copy=False)
 
-    @api.onchange('direction', 'fsm_order_id')
+    @api.onchange('warehouse_id', 'direction', 'fsm_order_id')
     def _onchange_location_id(self):
         super()._onchange_location_id()
         if self.fsm_order_id:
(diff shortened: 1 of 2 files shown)


fieldservice_stock/models/stock_request_order.py
```diff
@@ -12,8 +12,7 @@ class StockRequestOrder(models.Model):
         index=True, copy=False)
 
     @api.onchange('direction', 'fsm_order_id')
-    def _onchange_location_id(self):
-        super()._onchange_location_id()
+    def _onchange_direction_order_id(self):
```
patrickrwilson: Is this actually needing to be changed and the super() removed or just is it just needed to add the 'warehouse_id' to the onchange field list?

osi-scampbell: That was one of the old debugging pieces we had I am not sure how that got merged instead of what we finalized. Thanks for catching, just updated the PR

patrickrwilson (approved):

max3903 (approved):

PR #550, opened 2020-04-17
[ADD] fieldservice_stage_validation

brian10048:
As the stage of a field service document progresses, it may be important for specific fields of the document be completed.

For example, when a field service order enters the 'Complete' stage, the fields for 'Actual End Date' and/or 'Resolution Notes' should be set.

This module allows you to raise a Validation Error if the designated fields are not set when moving to a new stage. Since stages can be used for field service orders, workers, locations, or equipments, this logic is available to be applied to each document type.

@max3903 Let me know how this looks to you.  I'll add some tests to finish it off if everything looks good and update the settings in base fieldservice to allow the option to install it


fieldservice_stage_validation/models/fsm_stage.py
```diff
@@ -0,0 +1,31 @@
...
+        string='Fields to Validate',
+        help='Select fields which must be set on the document in this stage',
+    )
+
+    stage_type_model_id = fields.Many2one(
+        'ir.model',
+        compute=_compute_stage_model,
+        string='Model for Stage',
+        help='Technical field to hold model type',
+    )
```
max3903, line 31: @brian10048 Not sure you need an extra field. What about having a computed domain on validate_fields_ids?

fieldservice_stage_validation/models/fsm_equipment.py
```diff
@@ -0,0 +1,22 @@
...
+            stage = rec.stage_id
+            field_ids = stage.validate_field_ids
+            field_names = [x.name for x in field_ids]
+            values = rec.read(field_names)
+
+            for name in field_names:
+                if values[0][name] == False:
+                    raise ValidationError(_('Cannot move to stage "%s" ' \
+                                            'until the "%s" field is set.'
+                                            % (stage.name, name)))
```
max3903, line 22: @brian10048 Does it work without the loop and something like:

```
if any(values[0][field_names]) == False:
    raise ValidationError()
```

brian10048: @max3903 I get an `unhashable type: 'list'` error when trying `values[0][field_names]`

fieldservice_stage_validation/models/fsm_stage.py
brian10048: @max3903 I tried a handful of different options, but can't quite get the same result as I did here

max3903:
@brian10048 Any chance to get this ready to merge soon?

smangukiya:
@brian10048 : fieldservice_stage_validation/static/description/index.html is giving an error while installing this module, can you please take care of it. Other than that everything is working as expected.

brian10048:
@max3903 @smangukiya Everything here should be ready for merge pending your reviews

smangukiya (approved):

marcelsavegnago (approved):

PR #553, opened 2020-04-22
[IMP] fieldservice_vehicle_stock: Vehicle Required Bool + Mig Script

osi-scampbell:
The following PR allows users to select a boolean on Picking Types so they can decide which individual picking.types require a vehicle to be set on the picking before they can reserve quantities.

NOTE: I wasn't able to reference the the ID's in the migration script. 'env' is a cursor object and doesn't have attribute 'ref'. I updated based on name because if the user has multiple Vehicle Loadings then all of them should require a Vehicle being set.


fieldservice_vehicle_stock/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     'name': 'Field Service Vehicles - Stock',
     'summary': 'Track inventory of Field Service vehicles',
-    'version': '12.0.1.2.2',
+    'version': '12.0.2.2.1',
```
max3903:
```suggestion
    'version': '12.0.2.0.0',
```

fieldservice_vehicle_stock/models/stock_picking.py
```diff
@@ -11,26 +11,16 @@ class StockPicking(models.Model):
...
-                self.env.ref(
-                    'fieldservice_vehicle_stock.'
-                    'picking_type_location_to_vehicle')
-            ):
+            if rec.picking_type_id.require_vehicle_id:
                 if rec.fsm_vehicle_id:
                     picking = \
                         rec.with_context(vehicle_id=rec.fsm_vehicle_id.id)
-                    res = super(StockPicking, picking).action_assign()
+                    return super(StockPicking, picking).action_assign()
```
max3903: @osi-scampbell This will stop the loop and the logic will not apply to the next records...

fieldservice_vehicle_stock/data/fsm_stock_data.xml
```diff
@@ -153,6 +156,7 @@
                    ref="stock.stock_location_stock"/>
             <field name="return_picking_type_id"
                    ref="picking_type_output_to_vehicle"/>
+            <!-- <field name="require_vehicle_id" eval="True"/> -->
```
max3903: @osi-scampbell Comments can be removed.

fieldservice_vehicle_stock/migrations/12.0.1.2.1/post-migration.py
```diff
@@ -0,0 +1,13 @@
...
+
+def migrate(env, version):
+    if not version:
+            return
+
+    env.execute("""UPDATE stock_picking_type
+                SET require_vehicle_id = True
+                WHERE name in ('Vehicle Loading', 'Vehicle Returns',
+                'Location Pickup', 'Location Delivery');
+    """)
```
max3903: @osi-scampbell Directory needs to be renamed.

osi-scampbell: @max3903 when you say the directory needs to be renamed you are referring to the 12.0.1.2.1 directory? I thought it was supposed to be named after the previous version so any version after it will cause script to run?

fieldservice_vehicle_stock/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 {
     'name': 'Field Service Vehicles - Stock',
     'summary': 'Track inventory of Field Service vehicles',
-    'version': '12.0.1.2.2',
+    'version': '12.0.2.0.0',
```
dreispt, line 7: Don't do this, it will be doe my the merge bot.
You can add a comment to make sure maintainers use the `major` bump

fieldservice_vehicle_stock/models/stock_picking.py
```diff
@@ -11,26 +11,16 @@ class StockPicking(models.Model):
...
                 if rec.fsm_vehicle_id:
                     picking = \
                         rec.with_context(vehicle_id=rec.fsm_vehicle_id.id)
-                    res = super(StockPicking, picking).action_assign()
+                    res = super().action_assign()
                 else:
                     raise UserError(_(
                         "You must provide the vehicle for this picking type."))
-            res = super(StockPicking, rec).action_assign()
+            res = super().action_assign()
```
dreispt: We are calling call super() twice: on L19, and then again here.
Are you sure?

fieldservice_vehicle_stock/models/stock_picking_type.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class StockPickingType(models.Model):
+    _inherit = "stock.picking.type"
+
+    require_vehicle_id = fields.Boolean(
+        string='Require Vehicle',
+        help="Stock Pickings with this enabled must have the vehicle "
+             "set before they can reserve quantities. ",
```
dreispt: This explanation is not totally clear for me.

dreispt: " must have *this* vehicle set" maybe?

dreispt (commented):
Made some comments

fieldservice_vehicle_stock/__manifest__.py
max3903: @dreispt He needs to do this to determine the name of the migration directory.

fieldservice_vehicle_stock/models/stock_picking.py
osi-scampbell: @dreispt the second super() is for the if statement on line 15. If it does not go into that if statement then super() is not what we want.

fieldservice_vehicle_stock/__manifest__.py
osi-scampbell: @dreispt @max3903 I just reverted the version, I am going to bump it back

fieldservice_vehicle_stock/models/stock_picking_type.py
max3903: What about?
```
help="You must provide the vehicle to reserve those pickings."
```

dreispt:
@osi-scampbell Travis has error, seem to be unrelated. Can you rebase please?

PR #552, opened 2020-04-21
[IMP] fieldservice_route: Limit the number of day routes

max3903:

PR #554, opened 2020-04-22
[FIX] Day Route: recordset support for max check

dreispt:


max3903:
Merged in the branch of #552.

PR #556, opened 2020-04-27
[IMP] fsm_route: add function to get route schedule

dreispt:


max3903:
Merged in https://github.com/ursais/field-service/pull/30

PR #575, opened 2020-05-28
[IMP] fieldservice_isp_flow: ISP Stage Notifications

osi-scampbell:
In order for Notifications to be sent to followers of the message_subtypes we actually have to return those subtypes as seen in this PR.


max3903 (approved):

bodedra (approved):
LGTM :+1:

PR #584, opened 2020-07-08
feat: es_AR language

ibuioli:
Create translation file for es_AR language


marcelsavegnago:
Hello @ibuioli. Great job but translations must be sent via weblate. 

On the weblate platform it is possible to import the .po file

https://odoo-community.org/page/translate
https://docs.weblate.org/en/latest/user/translating.html

PR #589, opened 2020-07-10
[ADD] fieldservice_geo_route_optimization

brian10048:
This module provides a wizard that will sort and schedule orders in an optimized driving sequence

Depends on:
- [ ] https://github.com/OCA/geospatial/pull/264


wolfhall (commented):
Brian,
This is exciting to see progress on routing!  Thanks for pushing this forward.  
In our original design we have the concept of a day route for field workers.  The day route should include a starting point ( home, office, warehouse depot, etc.).  The day route has assigned jobs for the tech/driver and the day.  
The user can manually change the sequence, or call the "Route Sequence" optimization.(TSP).  I think your design allows the selection of orders but somehow this is not clear to me how this works when people want to combine manual and optimized sequencing.?
I suggest to use the day route structure.  this way people can manually change sequences.  this happens very often, e.g. if a customer is not available in the morning, etc.
If we store the OD pairs separately we can then also quickly recalculate the route based on travel time and provide realistic ETAs.

brian10048:
> 
> 
> Brian,
> This is exciting to see progress on routing! Thanks for pushing this forward.
> In our original design we have the concept of a day route for field workers. The day route should include a starting point ( home, office, warehouse depot, etc.). The day route has assigned jobs for the tech/driver and the day.
> The user can manually change the sequence, or call the "Route Sequence" optimization.(TSP). I think your design allows the selection of orders but somehow this is not clear to me how this works when people want to combine manual and optimized sequencing.?
> I suggest to use the day route structure. this way people can manually change sequences. this happens very often, e.g. if a customer is not available in the morning, etc.
> If we store the OD pairs separately we can then also quickly recalculate the route based on travel time and provide realistic ETAs.


A combination of manual and optimized sequencing is not necessarily implemented here and I agree it is something that needs to be available. I wanted to get something working first to start the discussion and figure we can always improve it next!

I agree that this could be tied into day routes more, but I also wondered if others would need this feature without requiring the day route module so I started with this simple wizard. It seems that the OCA prefers simple modules that add small features one at a time, rather than a more complete, rigid solution.

If the day route module is installed, using the wizard I created here should implicitly create a day route when the `scheduled_date_start` is updated and a worker is already assigned to the orders. A user could then make a manual change to the day route by updating a start time of an individual order in that day route. However, since I have not stored the distance matrix, making manual changes would not update the start time of the remaining orders without another API request to Google. Perhaps the addition of time windows would be the optimal solution. If a time window is put on a specific order, and then the optimization is ran, we should get the best sequence for a day route.

Day routes already offer fields to store the start and end location, a planned start time and maximum hours.  These are all parameters that the solver will accept to build an optimal route.  It would not be hard to introduce an Optimize button for the day route model that utilizes these fields to optimize its orders.

I'm curious if you have any models or examples of storing the OD pairs that would help me create them here?

Trying to maintain the idea of modules introducing one feature at a time, I wonder if this is the best route (pun intended) moving forward for field service optimization:

- `base_fieldservice_geo_optimization` - base module that inherits the geo optimization module from geospatial. Basically just pulling out the fsm_order model of this PR
- `fieldservice_geo_optimization_wizard` - split the wizard out of this module - fairly basic, provides an example of how to interact with the optimization model
- `fieldservice_route_geo_optimization` - module specifically for optimizing the orders of a day route, depends on the base module

Thoughts?

PR #593, opened 2020-07-17
[WIP][12.0][IMP] fieldservice

murtuzasaleh:
WIP

* Added a button next to the Worker filter on timeline view: "Synch Filters"
* If Clicked, and page is refreshed after Order filter changed, apply Order filter to Worker filter.
* In some cases the worker model may not have a matching field. in this case, don't set the filter.


murtuzasaleh:
@smangukiya Can you please review this PR?

PR #594, opened 2020-07-18
[13.0] [MIG] fieldservice_recurring

brian10048:
Migration to 13.0 for fieldservice_recurring

#354 

@max3903 @pedrobaeza In the process I mistakenly pushed to OCA/13.0 - Any chance one of you can fix that for me? Sorry! 😨


pedrobaeza:
Let's keep the branch with this module, and amend any problem in later PRs. Please take care with these questions. I wonder why do you have write access to the repo. Did you apply for PSC?

brian10048:
Sound good @pedrobaeza. I believe Max put me on the PSC as I have been active here on field-service since its beginning.  I'll be sure not to abuse any privileges 😄

pedrobaeza:
OK, no special problem, as you have contributed enough IMO.

PR #599, opened 2020-07-20
[13.0] [MIG] fieldservice_crm

brian10048:
Migration to 13.0 for fieldservice_crm

#354


patrickrwilson (approved):
LGTM
The .pot and .po files still reference version 12.0, i'm not familiar with what these are for but I assume they are auto generated files that will get updated? If not, they probably should be changed to 13.0.

brian10048:
Thank you! They are the translation files. I believe they get automatically changed to 13.0 once merged

bodedra (approved):
LGTM :+1:

PR #609, opened 2020-07-23
[FIX] fieldservice: fix timezone issue with Order filters

dreispt:
context_today does not work correctly for datetime filters,
it is being used durectly to compare wuth UTC stored datetimes.
To get the expected results, datetime.utcmow() must be used instead.

### DIFF
--- a/fieldservice/views/fsm_order.xml
+++ b/fieldservice/views/fsm_order.xml
@@ -298,19 +298,26 @@
                         name="done"/>
                 <separator/>
                 <filter string="Today Orders"
-                        domain="[('scheduled_date_start','&gt;=',context_today().strftime('%%Y-%%m-%%d 00:00:00')), ('scheduled_date_start','&lt;=',context_today().strftime('%%Y-%%m-%%d 23:59:59'))]"
-                        name="order_today"/>
+                    domain="[
+                    ('scheduled_date_start', '&gt;=', (datetime.datetime.combine(datetime.date.today(), datetime.time(0,0,0)))),
+                    ('scheduled_date_start', '&lt;', (datetime.datetime.combine(datetime.date.today(), datetime.time(23,59,59))))
+                    ]"
+                    name="order_today"/>
                 <filter string="Future Orders"
-                        domain="[('scheduled_date_start', '&gt;', context_today().strftime('%Y-%m-%d'))]"
-                        name="order_upcoming_all"/>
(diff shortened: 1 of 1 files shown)


smangukiya (approved):

smangukiya:
@dreispt : We need more approval to get this merge.  I can not set more reviewers to this so please do that from your end. Thanks

brian10048:
I will be glad to test and provide a review soon

brian10048:
I don't think this solves the issue.

I am testing on the runbot build.
I created an order for todays date at 00:00:01
I applied the Today filter
No orders show in the view.

I then made several attempts to the order schedule start time.  For me, the cutoff seems to be at 20:00:00.  An order with time of 19:59:59 (with today's date) will not show in today's order filter.  However, 20:00:00 and past show in the Today filter.  Similarly, I put another order for tomorrow's date. Anytime before 20:00:00 tomorrow will show up in the Today's filter, even though it is on tomorrow's date.

I tried changing the User Preferences time zone.  Changes to the time zone had no effect on the result.  My computer is set to timezone of US/Eastern which is 4 hours from UTC and equivalent to the difference (00:00:00 to 20:00:00)

brian10048:
Testing on my current database to reproduce the bug (without this PR):

An order set for yesterday's date at 20:00:00 will show in 'Today's' filter option.  An order will not show in the Today filter if it's scheduled start date is after today's date at 19:59:59

Once again, the offset for me locally is 4 hours from UTC

brian10048 (changes requested):
https://github.com/odoo/odoo/blob/f7567f10dd0cc936336066e7835c908fde163dc0/addons/stock_landed_costs/views/stock_landed_cost_views.xml#L193:199

Similar concept from Odoo code uses context_today, but without times

osi-scampbell:
@brian10048 @dreispt @max3903 Hey Team, I was able to reproduce this issue using runbot here:
image
Observe when the filter is applied we cannot see our FSM Order for today (08/13)
image

After adding this filter through the front end "Edit Form View", this is the results
image

I believe this is the proper solution to this issue.

marcelsavegnago (approved):

brian10048 (approved):
Tested on runbot and functions as intended.

Thank you!

As a side note, should we update the filter names so we don't have 3 with the same name "order_upcoming_all"?

brian10048:
@dreispt @osi-scampbell Can you review/merge https://github.com/ursais/field-service/pull/34 and then I'm good with merging this one to OCA

bodedra (approved):

PR #614, opened 2020-08-17
[FIX] fieldservice_distribution (Smart Button Count)

patrickrwilson:
This fixes the functionality of the 'Distributed To' smart button. Before this fix, the smart button was looking for locations that had it's fsm_parent_id set which is incorrect because we want to link locations that have their parent distribution set to the record. 

With this fix, the 'Distributed To' smart button now counts all locations that have the 'Parent Distribution' set to the current record, and it then opens the views with those proper filter. It also shows only direct relationships as distribution instead of all children since we only care about direct child unlike the sub location smart button.

### DIFF
--- a/fieldservice_distribution/models/fsm_location.py
+++ b/fieldservice_distribution/models/fsm_location.py
@@ -19,7 +19,7 @@ class FSMLocation(models.Model):
     def _compute_distrib_sublocation_ids(self):
         for location in self:
             location.distrib_count = self.env['fsm.location'].search_count(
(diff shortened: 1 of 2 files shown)


patrickrwilson:
Ping @wolfhall @max3903 @osi-scampbell

osi-scampbell:
@patrickrwilson This PR looks good to me. Approved.

max3903 (approved):

marcelsavegnago (approved):

PR #615, opened 2020-08-20
[IMP] fieldservice: Give Access to portal user

murtuzasaleh:

PR #618, opened 2020-08-20
[12.0][IMP] fieldservice_account:- Gave Access to portal user.

murtuzasaleh:


bodedra (approved):

marcelsavegnago (approved):

fieldservice_account/security/ir.model.access.csv
```diff
@@ -1,2 +1,6 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_account_invoice_line_fsm_user,account.invoice.line.fsm.user,model_account_invoice_line,fieldservice.group_fsm_user,1,1,1,1
+access_account_invoice_line_portal,access.account.invoice.line.portal,account.model_account_invoice_line,base.group_portal,1,0,0,0
```
brian10048, line 3: This is already in [odoo accounting](https://github.com/odoo/odoo/blob/12.0/addons/account/security/ir.model.access.csv)

fieldservice_account/security/ir.model.access.csv
```diff
@@ -1,2 +1,6 @@
...
+access_account_account_portal,access.account.account.portal,account.model_account_account,base.group_portal,1,0,0,0
+access_account_move_portal,access.account.move.portal,account.model_account_move,base.group_portal,1,0,0,0
+access_account_move_line_portal,access.account.move.line.portal,account.model_account_move_line,base.group_portal,1,0,0,0
```
brian10048, line 6: I'm curious why we need to grant these permissions for these models from accounting?

dreispt:
@bodedra Are you sure these changes are needed? They seem redundant.

fieldservice_account/security/ir.model.access.csv
dreispt: @bodedra I have security concerns with this - it makes the Accounting moves an open book for any Portal user. I would vote against adding this.

fieldservice_account/security/ir.model.access.csv
dreispt: I [confirm](https://github.com/odoo/odoo/blob/12.0/addons/account/security/ir.model.access.csv#L43) that. @bodedra ?

dreispt (changes requested):
I have concerns with these changes. Can we have clarification?

fieldservice_account/security/ir.model.access.csv
bodedra: Good catch @brian10048  This changes are redundant.

fieldservice_account/security/ir.model.access.csv
bodedra: @dreispt I agree with you. @max3903 @murtuzasaleh would you please provide your input?

bodedra (changes requested):
Based on recent comments, I down vote for now.

marcelsavegnago:
> Based on recent comments, I down vote for now.

me too

fieldservice_account/security/ir.model.access.csv
brian10048: https://github.com/OCA/field-service/pull/615#discussion_r476673789 - See similar issue and responses here.  Considering these changes are for the mobile app, shouldn't they go in the repo for the mobile app? (which I'm assuming is not going to be OCA)

wolfhall: We need a way for portal users (via website or mobile) to access and update FSM statuses.  We should not need to give them access to invoice lines or other accounting info.  So if a portal user uses FSM, there will be limitations to what they can do.  Because the goal here is simple web-portal access for contractors it is not FSM-mobile specific (otherwise it should not be in OCA).  @murtuzasaleh - we nee to define what portal users will not need to access before we drill holes security.

murtuzasaleh:
@dreispt @marcelsavegnago @brian10048 @bodedra I will close this PR because we have handle access right issue from the mobile side (Improved code).
Thanks

PR #629, opened 2020-08-24
[12.0][IMP] fieldservice : Use commercial_partner_id when convert a fsm_location

hparfr:
Because it makes more sense when a child contact of company is converted to keep the owning to the parent company and it makes no difference when there is no parent_id relation in the partner.

### DIFF
--- a/fieldservice/wizard/fsm_wizard.py
+++ b/fieldservice/wizard/fsm_wizard.py
@@ -31,7 +31,7 @@ class FSMWizard(models.TransientModel):
     def _prepare_fsm_location(self, partner):
         return {
             'partner_id': partner.id,
-            'owner_id': partner.id,
+            'owner_id': partner.commercial_partner_id.id
         }
 
     def action_convert_location(self, partner):


fieldservice/wizard/fsm_wizard.py
```diff
@@ -31,7 +31,7 @@ def action_convert(self):
     def _prepare_fsm_location(self, partner):
         return {
             'partner_id': partner.id,
-            'owner_id': partner.id,
+            'owner_id': partner.commercial_partner_id.id
```
dreispt, line 34: Iv've seen cases where the we have multiple levels here. I would rather use `parent_id`, if possible.
```suggestion
            'owner_id': partner.parent_id.id
```

hparfr: What is the hierarchy of partners you have ?
I don't think odoo supports a multi level : a company can't really be a child of another company without side effects.

dreispt:
Not out of the box, but it can be made to work. Example is using a structure like Company -> Branch/Store -> Contact.

I’ve also seen Group Company -> Company -> Contact. 

--dr

PR #636, opened 2020-09-10
[FIX] FSM Invoice Double Timesheet Entry

osi-scampbell:
The Contractor Costs field is a One2many relationship between fsm.orders and account.invoice.lines. By looping through the Employee Timesheet entries first, we generate new rows in this table which are then picking up a 2nd time when we loop through the contractor costs. By looping through the Contractor costs first, and then the Employee Timesheets, these lines are not duplicated on our invoice.


marcelsavegnago (approved):

dreispt (approved):

bodedra (approved):
/ocabot merge patch

PR #643, opened 2020-09-25
[12.0][FIX] Improved code of stage create method.

rutul-serpentcs:
When I create stage from Field Service >> Dashboard >> Team >> Orders kanban view I got an error.

Teams-Odoo

### DIFF
--- a/fieldservice/models/fsm_stage.py
+++ b/fieldservice/models/fsm_stage.py
@@ -45,8 +45,7 @@ class FSMStage(models.Model):
     stage_type = fields.Selection([('order', 'Order'),
                                    ('equipment', 'Equipment'),
                                    ('location', 'Location'),
-                                   ('worker', 'Worker')], 'Type',
-                                  required=True)
+                                   ('worker', 'Worker')], 'Type', require=True)
     company_id = fields.Many2one(
         'res.company', string='Company',
         default=lambda self: self.env.user.company_id.id)
@@ -73,8 +72,8 @@ class FSMStage(models.Model):
     def create(self, vals):
         stages = self.env['fsm.stage'].search([])
         for stage in stages:
-            if stage.stage_type == vals['stage_type'] and \
-               stage.sequence == vals['sequence']:
+            if stage.stage_type == vals.get('stage_type') and \
+               stage.sequence == vals.get('sequence'):
                 raise ValidationError(_("Cannot create FSM Stage because "
                                         "it has the same Type and Sequence "
                                         "of an existing FSM Stage."))
--- a/fieldservice/views/fsm_team.xml
+++ b/fieldservice/views/fsm_team.xml
@@ -13,7 +13,8 @@
         <field name="name">Orders</field>
         <field name="res_model">fsm.order</field>
         <field name="view_mode">kanban,timeline,tree,form,calendar</field>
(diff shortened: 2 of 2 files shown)


brian10048 (changes requested):
@rutul-serpentcs 
This fixes the original error you presented.
However there is still an issue creating the stage from Kanban view

```
The operation cannot be completed:
- Create/update: a mandatory field is not set.
- Delete: another model requires the record being deleted. If possible, archive it instead.

Model: Field Service Stage (fsm.stage), Field: Type (stage_type) 
```
stage_type field needs to be set

rutul-serpentcs:
> @rutul-serpentcs
> This fixes the original error you presented.
> However there is still an issue creating the stage from Kanban view
> 
> ```
> The operation cannot be completed:
> - Create/update: a mandatory field is not set.
> - Delete: another model requires the record being deleted. If possible, archive it instead.
> 
> Model: Field Service Stage (fsm.stage), Field: Type (stage_type) 
> ```
> 
> stage_type field needs to be set

@brian10048 Done, Can you please review it again?

brian10048:
@rutul-serpentcs - It allows to create the stage, but a new bug is introduced when we create a stage that has an undefined stage type.  When you go back to the orders kanban view, the stage recently created will not be displayed.

Can you revert the recent change and keep the stage_type required in python?  Then, instead can you use the context to set the correct `stage_type` based on the view

fieldservice/models/fsm_stage.py
```diff
@@ -45,8 +45,7 @@ def _default_team_ids(self):
     stage_type = fields.Selection([('order', 'Order'),
                                    ('equipment', 'Equipment'),
                                    ('location', 'Location'),
-                                   ('worker', 'Worker')], 'Type',
-                                  required=True)
+                                   ('worker', 'Worker')], 'Type', require=True)
```
dreispt, line 48: I'm afraid this is incorrect,  `required` is the correct field attribute.
Is the goal to remove the required attribute?

fieldservice/views/fsm_team.xml
```diff
@@ -13,7 +13,8 @@
         <field name="name">Orders</field>
         <field name="res_model">fsm.order</field>
         <field name="view_mode">kanban,timeline,tree,form,calendar</field>
-        <field name="context">{'default_team_id': active_id}</field>
+        <field name="context">{'default_team_id': active_id,
+            'default_stage_type':'order'}</field>
```
dreispt, line 17:
```suggestion
            'default_stage_type': 'order'}</field>
```

fieldservice/models/fsm_stage.py
```diff
@@ -73,8 +72,8 @@ def get_color_information(self):
     def create(self, vals):
         stages = self.env['fsm.stage'].search([])
         for stage in stages:
-            if stage.stage_type == vals['stage_type'] and \
-               stage.sequence == vals['sequence']:
+            if stage.stage_type == vals.get('stage_type') and \
+               stage.sequence == vals.get('sequence'):
```
dreispt, line 76: :+1:

max3903 (changes requested):
Please attend the comments

ivantodorovich:
I think we can close this one (or remove the "no stale" label), given it's old and the comments haven't been taken care of.

PR #644, opened 2020-09-25
[12.0][IMP] Added check for existing fsm.equipment before creating

barnarj:
Added a check to verify if a related fsm.equipment object exists before attempting to create one for assigning to the line.lot_id.fsm_equipment_id field.

This mitigates an issue we encountered during outgoing stock moves involving equipment which had previously been deployed to an fsm.location and then recovered.


dreispt (approved):
LGTM

dreispt:
@osi-scampbell @max3903 I would rather have one if you merge this one, please.

PR #651, opened 2020-10-06
[13.0] [MIG] fieldservice_stock

RLeeGML:
Migration of fieldservice_stock to 13.0 

#354 

Depends on [stock-logistics-warehouse](https://github.com/OCA/stock-logistics-warehouse):

- [x] [#865](https://github.com/OCA/stock-logistics-warehouse/pull/865) 
- [x] [#867](https://github.com/OCA/stock-logistics-warehouse/pull/867)


bodedra:
@RLeeOSI can you run "pre-commit run -a" command to ensure travis-ci/pr green?

RLeeGML:
@bodedra The problem is that black is editing stock_move.py and making 2 lines too long. Then flake8 catches the lines being too long and kicks up the error. If we try to fix it, black will just reverse what we do and flake8 will get the error again. Is there a way to avoid this, or to force black to skip those lines?

RLeeGML:
Closed for #671

PR #602, opened 2020-07-21
[13.0] [MIG] fieldservice_skill

brian10048:
Migrate fieldservice_skill to 13.0

For #354

PR #658, opened 2020-10-09
[13.0] [MIG] fieldservice_skill

RLeeGML:
Migration of fieldservice_skill to 13.0

#354


RLeeGML:
@max3903 @osi-scampbell 

The dependency hr_skill from OCA/hr has been changed to hr_skills from odoo/odoo. The old module had a color field on hr.skill model, while the new one does not. Does this need to be added back in?

The Usage file indicates that the worker list on fsm.order will be filtered by worker skill (as well as location) but there's no functionality to filter by worker skill in the v12 module. Should this be added in?

RLeeGML:
PR for this module already exists
#602

PR #663, opened 2020-10-29
[IMP] fieldservice: basic security

chienandalu:
- Add a lower security group whose members only have access to their own
field service orders.
- Add related partner to location and worker forms to easily check the
relationship.

cc @Tecnativa TT25235


fieldservice/views/res_partner.xml
```diff
@@ -29,7 +29,7 @@
                         name="action_open_owned_locations"
                         icon="fa-map-marker"
                         context="{'default_owner_id': active_id}"
-                        groups="fieldservice.group_fsm_user">
+                        groups="fieldservice.group_fsm_user_own">
```
pedrobaeza, line 32: This is not needed having the group at view level

pedrobaeza (approved):

brian10048 (approved):

PR #668, opened 2020-11-10
[12.0][IMP] fieldservice_sale: FSM Sale Template Team

osi-scampbell:
Current Behavior in runbot v12:
Product with FSM Teamplate on it and the FSM Template has a Team:
image
image
image

Upon confirming the Sales Order, an FSM Order is generated but the neither the Template, nor the Team is brought over with it. 
image

This PR corrects that behavior.


brian10048:
@osi-scampbell 
I think we should also change [lines 58-61 here](https://github.com/OCA/field-service/pull/668/commits/a31556bd7f56a3dbcb866815cf5799fdb308b33f#diff-6e8936d0138c6bf3bd522b7dae70fce0a65e1120d0f7aeed93ea6593ab9a41f3R58-R61).

If 2 different products are sold with different FSM templates, only 1 FSM order for the first template in the sale order is created.  I think it should create 1 FSM order for each FSM template in the sale order.  

In a scenario where there are 5 unique products with FSM template A and 3 unique products with FSM template B on one sale order, we would get 1 FSM order with Template A and 1 FSM order with Template B.  Currently we only get 1 FSM order with template A

bodedra:
@osi-scampbell may you add comment from @brian10048

osi-scampbell:
> @osi-scampbell
> I think we should also change [lines 58-61 here](https://github.com/OCA/field-service/pull/668/commits/a31556bd7f56a3dbcb866815cf5799fdb308b33f#diff-6e8936d0138c6bf3bd522b7dae70fce0a65e1120d0f7aeed93ea6593ab9a41f3R58-R61).
> 
> If 2 different products are sold with different FSM templates, only 1 FSM order for the first template in the sale order is created. I think it should create 1 FSM order for each FSM template in the sale order.
> 
> In a scenario where there are 5 unique products with FSM template A and 3 unique products with FSM template B on one sale order, we would get 1 FSM order with Template A and 1 FSM order with Template B. Currently we only get 1 FSM order with template A

@brian10048 Are you referring to this line of code here: https://github.com/ursais/field-service/blob/a31556bd7f56a3dbcb866815cf5799fdb308b33f/fieldservice_sale/models/sale_order.py#L58-L61 ? 

If so, I think you are correct. It looks like it gets all of the tags, instructions, and hours together from all the templates and adds them into 1 FSM Order. You are requesting that these stay separate. Reviewing the code, at line 53 of sale_order.py they map all the lines who's product_id.field_service_tracking == 'sale' (1 FSM Order per Sale Order). I don't see anything to handle if product_id.field_service_tracking == 'line' (1 FSM Order per Sale Order Line. I think that is the cause of this issue, what do you think?

bodedra:
@brian10048 any feedback on @osi-scampbell response?

PR #669, opened 2020-11-11
[12.0][IMP] fieldservice_maintenance: Convert Maintenance Equipment To FSM

osi-scampbell:
The following PR adds a wizard that allows users to convert a Maintenance Equipment into an FSM Equipment similar to how we convert Contacts into FSM Records

### DIFF
--- a/fieldservice_maintenance/__manifest__.py
+++ b/fieldservice_maintenance/__manifest__.py
@@ -18,6 +18,7 @@
         'views/maintenance_request_view.xml',
         'views/fsm_equipment_view.xml',
         'views/fsm_order_view.xml',
+        'wizard/maintenance_equipment_wizard.xml'
     ],
     'pre_init_hook': 'pre_init_hook',
     'license': 'AGPL-3',
--- a/fieldservice_maintenance/__manifest__.py
+++ b/fieldservice_maintenance/__manifest__.py
@@ -14,10 +14,12 @@
     ],
     'data': [
         'data/fsm_order_type.xml',
+        'security/ir.model.access.csv',
         'views/maintenance_equipment_view.xml',
         'views/maintenance_request_view.xml',
         'views/fsm_equipment_view.xml',
         'views/fsm_order_view.xml',
+        'wizard/maintenance_equipment_wizard.xml'
     ],
     'pre_init_hook': 'pre_init_hook',
     'license': 'AGPL-3',
--- a/fieldservice_maintenance/__manifest__.py
+++ b/fieldservice_maintenance/__manifest__.py
@@ -14,7 +14,6 @@
     ],
     'data': [
         'data/fsm_order_type.xml',
-        'security/ir.model.access.csv',
         'views/maintenance_equipment_view.xml',
         'views/maintenance_request_view.xml',
         'views/fsm_equipment_view.xml',
--- a/fieldservice_maintenance/models/maintenance_equipment.py
+++ b/fieldservice_maintenance/models/maintenance_equipment.py
@@ -1,10 +1,11 @@
 # Copyright (C) 2018 Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class MaintenanceEquipment(models.Model):
     _inherit = 'maintenance.equipment'
 
     is_fsm_equipment = fields.Boolean(string='Is a FSM Equipment')
+    product_id = fields.Many2one('product.product', 'Product')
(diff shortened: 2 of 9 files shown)


fieldservice_maintenance/wizard/maintenance_equipment_wizard.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (C) 2018 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class MainenanceEquipmentWizard(models.Model):
```
bodedra:
```suggestion
class MainenanceEquipmentWizard(models.TransientModel):
```

fieldservice_maintenance/__manifest__.py
```diff
@@ -14,10 +14,12 @@
     ],
     'data': [
         'data/fsm_order_type.xml',
+        'security/ir.model.access.csv',
```
bodedra:
```suggestion
```

bodedra (changes requested):
Few points to improves:

- wizard should be `TransientModel`
- remove security file.
- create a lot number if product has tracking = serial or lot.

fieldservice_maintenance/wizard/maintenance_equipment_wizard.py
```diff
@@ -0,0 +1,43 @@
...
+    _name = 'maintenance.equipment.wizard'
+
+    @api.multi
+    def convert_maintenance_equipment_fsm(self):
+        maintenance_equpment_ids = self.env['maintenance.equipment'].\
+            browse(self._context.get('active_ids', []))
+        for maintenance_id in maintenance_equpment_ids:
+            if not maintenance_id.serial_no or not maintenance_id.serial_no:
```
bodedra: `maintenance_id.serial_no` condition appear twice in a condition.

fieldservice_maintenance/models/maintenance_equipment.py
```diff
@@ -1,10 +1,11 @@
 # Copyright (C) 2018 Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
```
bodedra, line 4:
```suggestion
from odoo import fields, models
```

bodedra (changes requested):
@osi-scampbell 

add this changes and squash your commits.

Once this is green, I will review it.

bodedra (changes requested):
Functional review :+1:

bodedra:
@brian10048 @hparfr @marcelsavegnago Are you guys available to review this?

bodedra (approved):

fieldservice_maintenance/wizard/maintenance_equipment_wizard.xml
```diff
@@ -0,0 +1,29 @@
...
+                target="new"
+                key2="client_action_multi"
+                groups="fieldservice.group_fsm_dispatcher"/>
+
+    <!-- wizard view -->
+    <record id="maintenance_fsm_wizard_view" model="ir.ui.view">
+        <field name="name">Convert to a FSM Equipment</field>
+        <field name="model">maintenance.equipment.wizard</field>
+        <field name="arch" type="xml">
+            <form string="Convert to a Field Service Worker or Location">
```
hparfr: wrong string

hparfr (approved):
I'm not familiar with fsm_equipements so I don't know what are the benefits of the "convert to..."

If you want to align with fsm_location you will have to "inherits" https://github.com/OCA/field-service/blob/12.0/fieldservice/models/fsm_location.py#L10

Are you sure adding product_id should be part of this PR ?

marcelsavegnago (approved):

osi-scampbell:
> I'm not familiar with fsm_equipements so I don't know what are the benefits of the "convert to..."
> 
> If you want to align with fsm_location you will have to "inherits" https://github.com/OCA/field-service/blob/12.0/fieldservice/models/fsm_location.py#L10
> 
> Are you sure adding product_id should be part of this PR ?

When an FSM Equipment is created, a Maintenance Equipment is automatically created. However when you create a Maintenance Equipment, an FSM Equipment is not automatically created. So this lets users take a maintenance equipment that has already been created and convert it to an FSM Equipment. 

You bring up a very good point about the product_id addition, I have had others ask the same question. I added this because on an FSM Equipment, product_id and lot_id are required fields. Perhaps this shouldn't be the case considering they aren't needed on Maintenance Equipment so maybe they shouldn't be needed on FSM Equipment? I will follow up on this question today and get back to you, I am not sure what the ramifications would be if we made product_id and lot_id no longer required on FSM Equipment

dreispt:
https://travis-ci.com/github/OCA/field-service/jobs/450063553#L602

osi-scampbell:
@dreispt It looks like these are lint errors in JavaScript that are in fieldservie and in fieldservice_vehicle_stock. Should those be fixed in this PR or should they be in a separate PR since they are in separate modules?

marcelsavegnago:
ping @dreispt

PR #657, opened 2020-10-09
[13.0] [MIG] fieldservice_maintenance

RLeeGML:
Migration of fieldservice_maintenance to 13.0

#354 

Depends on commit [8e5a34e](https://github.com/OCA/field-service/commits/8e5a34e1fd1b1317b723ae6ddd74e727d7b98f85) -- commit made to 12.0 after migration of fieldservice to 13.0

PR #670, opened 2020-11-18
[13.0] [MIG] fieldservice_maintenance

RLeeGML:
Migration of fieldservice_maintenance to 13.0

#354 

Moved from #657


fieldservice_maintenance/models/fsm_equipment.py
```diff
@@ -0,0 +1,51 @@
...
+    _inherit = "fsm.equipment"
+
+    maintenance_equipment_id = fields.Many2one(
+        "maintenance.equipment",
+        string="Related Maintenance Equipment",
+        readonly=True,
+        ondelete="restrict",
+        delegate=True,
+        auto_join=True,
+        index=True,
```
apineux, line 17: You need required=True for delegate fields.
If the reason is that the field becomes required on the Form view, you can adapt the attrs required with False in the view.

fieldservice_maintenance/models/fsm_equipment.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def unlink(self):
+        equipments = self.mapped("maintenance_equipment_id")
+        res = super(FSMEquipment, self).unlink()
+        for equipment in equipments:
+            other = self.env["fsm.equipment"].search(
+                [("maintenance_equipment_id", "=", equipment.id)]
+            )
+            if not other:
+                equipment.is_fsm_equipment = False
```
apineux, line 39: is_fsm_equipment should be a computed field. If you are using a One2many fsm_equipment_ids on maintenance.equipment then your computed field just have to look if fsm_equipment_ids contains records.
Therefore you don't have to overwrite the unlink anymore.

fieldservice_maintenance/models/fsm_equipment.py
```diff
@@ -0,0 +1,51 @@
...
+        auto_join=True,
+        index=True,
+    )
+
+    @api.model
+    def create(self, vals):
+        maintenance_equipment_id = self.env["maintenance.equipment"].create(
+            self._prepare_maintenance_vals(vals)
+        )
+        if maintenance_equipment_id:
```
apineux, line 25: The "if" here is not necessary because if the create doesn't work, you'll have an exception

fieldservice_maintenance/models/fsm_equipment.py
```diff
@@ -0,0 +1,51 @@
...
+            vals.update({"maintenance_equipment_id": maintenance_equipment_id.id})
+        res = super().create(vals)
+        maintenance_equipment_id.fsm_equipment_id = res.id
```
apineux, line 28: You should use a one2many (fsm_equipment_ids) on maintenance.equipment model. Therefore no need to write on the maintenance_equipment after the create.

fieldservice_maintenance/models/maintenance_equipment.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MaintenanceEquipment(models.Model):
+    _inherit = "maintenance.equipment"
+
+    is_fsm_equipment = fields.Boolean(string="Is a FSM Equipment")
+    fsm_equipment_id = fields.Many2one("fsm.equipment", "FSM Equipment", readonly=True)
```
apineux, line 11: This should be a one2many. Is there any reason why you use a many2one instead?

fieldservice_maintenance/models/maintenance_equipment.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (C) 2018 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MaintenanceEquipment(models.Model):
+    _inherit = "maintenance.equipment"
+
+    is_fsm_equipment = fields.Boolean(string="Is a FSM Equipment")
```
apineux, line 10: This should be computed field.

fieldservice_maintenance/models/maintenance_request.py
```diff
@@ -0,0 +1,40 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class MaintenanceRequest(models.Model):
+    _inherit = "maintenance.request"
+
+    fsm_order_id = fields.Many2one("fsm.order", "Field Service Order")
```
apineux, line 11: Why don't use a one2many fields with inverse_name = request_id ?
With your current implementation when you create the maintenance.request from the fsm.order your fsm_order_id is empty. 
You should have a same result from both side therefore it is better to use a one2many.

If you need a "one2one" relation I think it is better to apply a constrains()

apineux (changes requested):
Hello, thank you for your work.

I added some comments with questions and suggestions, thanks in advance for taking a look.

Regards,

fieldservice_maintenance/wizard/maintenance_equipment_wizard.py
```diff
@@ -0,0 +1,38 @@
...
+class MainenanceEquipmentWizard(models.TransientModel):
+    _name = "maintenance.equipment.wizard"
+    _description = "Mainenance Equipment Wizard"
+
+    def convert_maintenance_equipment_fsm(self):
+        maintenance_equpment_ids = self.env["maintenance.equipment"].browse(
+            self._context.get("active_ids", [])
+        )
+        for maintenance_id in maintenance_equpment_ids:
+            if not (maintenance_id.product_id or maintenance_id.lot_id):
```
baimont, line 17: Hi there! Thanks for this pr.
I have a problem with this line.
It leads to a bug since there is no product_id on a maintenance.equipment.

PR #678, opened 2020-12-11
[13.0][FIX] fieldservice: Schedule Date End Bug

osi-scampbell:
There is bug related to the calculation of scheduled_date_end, here is the bug reproduced in runbot. Test cases to be added for this part of the code.
image
image
image

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -166,7 +166,9 @@ class FSMOrder(models.Model):
     scheduled_duration = fields.Float(
         string="Scheduled duration", help="Scheduled duration of the work in" " hours"
     )
-    scheduled_date_end = fields.Datetime(string="Scheduled End")
+    scheduled_date_end = fields.Datetime(string="Scheduled End",
+                                         readonly=True,
+                                         compute="_calc_scheduled_dates")
     sequence = fields.Integer(string="Sequence", default=10)
     todo = fields.Text(string="Instructions")
 
@@ -259,7 +261,7 @@ class FSMOrder(models.Model):
                 or False
             }
         )
-        self._calc_scheduled_dates(vals)
+        self._calc_scheduled_dates()
         if not vals.get("request_late"):
             if vals.get("priority") == "0":
                 if vals.get("request_early"):
@@ -291,7 +293,6 @@ class FSMOrder(models.Model):
             stage_id = self.env["fsm.stage"].browse(vals.get("stage_id"))
             if stage_id == self.env.ref("fieldservice.fsm_stage_completed"):
                 raise UserError(_("Cannot move to completed from Kanban"))
-        self._calc_scheduled_dates(vals)
         res = super(FSMOrder, self).write(vals)
         return res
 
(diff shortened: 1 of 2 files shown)


bodedra (approved):
Functional review. :+1:

bodedra:
@brian10048 @hparfr @marcelsavegnago

hparfr (changes requested):
It works. Thanks.


Because there is: ` <field name="scheduled_date_end" readonly="1" /> `, scheduled_date_end can't be modified so it's never in vals.

I suggest to remove the feature to set a scheduled_date_end. It will simplify the  `def _calc_scheduled_dates(self, vals):`
Change : 
```py
     scheduled_date_end = fields.Datetime(string="Scheduled End")
```

to:
```py 
   scheduled_date_end = fields.Datetime(string="Scheduled End", readonly=True, compute="_calc_scheduled_dates")
```

And add @pi.depends
```py
   @api.depends("scheduled_date_start", "scheduled_date_duration")
    def _calc_scheduled_dates(self, vals):


```

bodedra:
@osi-scampbell Please address comment from @hparfr

osi-scampbell:
@hparfr @bodedra Thank you for the feedback, I have updated the PR and will be adding test cases as well.

bodedra:
I will review it once runbot is ready. Thank you @osi-scampbell

bodedra:
@hparfr @brian10048  @max3903 @dreispt  Can you review this?

fieldservice/models/fsm_order.py
```diff
@@ -306,52 +307,40 @@ def unlink(self):
...
-            elif vals.get("scheduled_duration", False) or (
-                vals.get("scheduled_date_start", False)
-                and (
-                    self.scheduled_date_start != vals.get("scheduled_date_start", False)
-                )
+        for order in self:
+            if (
+                order.scheduled_duration
+                or order.scheduled_date_start
+                or order.scheduled_date_end
```
hparfr, line 317: Something like this will be enough.
```suggestion
       
        for order in self:
            if order.scheduled_duration and order.scheduled_date_start:
                hours = order.scheduled_duration
                order.scheduled_date_end = order.scheduled_date_start + timedelta(
                        hours=hours
                    )
            else:
                order.scheduled_date_end = False;
```

fieldservice/models/fsm_order.py
```diff
@@ -306,52 +307,40 @@ def unlink(self):
...
+
+                elif order.scheduled_duration or (
+                    order.scheduled_date_start
+                    and (self.scheduled_date_start != order.scheduled_date_start)
+                ):
+                    hours = order.scheduled_duration
+                    start_date_val = order.scheduled_date_start
+                    start_date = fields.Datetime.from_string(start_date_val)
+                    date_to_with_delta = start_date + timedelta(hours=hours)
+                    order.scheduled_date_end = str(date_to_with_delta)
```
hparfr, line 343: To be removed

hparfr (changes requested):

PR #686, opened 2021-01-12
[WIP][12.0][IMP] Split Method For Computed Fields

osi-scampbell:
Split methods to improve run time


hparfr:
Hi,

Can you remove `res.partner[service_location_id]` ? It's never assigned.

PR #725, opened 2021-02-12
[14.0][MIG] fieldservice_sale

Freni-OSI:


fieldservice_sale/models/sale_order.py
```diff
@@ -0,0 +1,213 @@
...
+        lines = self.order_line.filtered(
+            lambda sol: sol.product_id.field_service_tracking == "sale"
+        )
+        templates = lines.mapped("product_id.fsm_order_template_id")
+        note = ""
+        hours = 0.0
+        categories = self.env["fsm.category"]
+        for template in templates:
+            note += template.instructions or ""
+            hours += template.hours
```
osi-scampbell, line 67: @Freni-OSI This should be template.duration

osi-scampbell (changes requested):

fieldservice_sale/models/sale_order.py
Freni-OSI: @osi-scampbell Changed in Latest PR #761.

PR #741, opened 2021-03-12
[FIX][12.0] fsm.order unlink method only evaluates first record

HekkiMelody:
This doesn't cause issues when deleting a single order, but when selecting multiple orders (e.g. from the list view), it will only evaluate the first one. 

WIth this change, all records are evaluated and deleted, unless the ValidationError is raised.

Just looking at the code, the 13.0 version seems to have the same issue, though I haven't tested it.

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -260,12 +260,10 @@ class FSMOrder(models.Model):
 
     @api.multi
     def unlink(self):
-        for order in self:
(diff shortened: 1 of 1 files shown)


hparfr:
In v13, v14, it delete only the first one even if there is no error raised.

brian10048 (approved):
Tested on runbot.  Successfully deleted multiple records

marcelsavegnago (approved):
LGTM

PR #743, opened 2021-03-15
[FIX][12.0] fieldservice_status (Field Requirement)

patrickrwilson:
There was an issue where the sub status field was hidden on the form for anything other than FSM Orders however it was also required. This PR just updates the attributes to only require the field when the stage is an FSM order.


patrickrwilson:
@osi-scampbell can you review this PR?

bodedra (approved):
LG :+1:

marcelsavegnago (approved):

PR #751, opened 2021-03-26
[14.0][FIX] fieldservice: remove ref from fsm.location

hparfr:
because ref is already available from res.partner

With this fix, it's less error prone, aligned with other fields like street, phone, ... and it works nicely with other oca modules from partner-contact repo like partner_ref_unique


mourad-ehm (approved):

marcelsavegnago (approved):
LGTM

brian10048 (approved):
/ocabot merge patch

PR #650, opened 2020-10-06
[13.0][MIG] fieldservice_account_analytic

murtuzasaleh:
Depends on #598

PR #773, opened 2021-04-28
[13.0][MIG] fieldservice_account_analytic

baimont:
https://github.com/OCA/field-service/issues/354

Hi there,
I know @murtuzasaleh already did a fine job there https://github.com/OCA/field-service/pull/650.
But I was trying to migrate fieldservice_isp_account which depends on fieldservice_account_analytic... And I had a problem with unit tests.
Tests on https://github.com/OCA/field-service/pull/650 are not working because they depend on fieldservice_account's setUp() which creates test_location without customer_id. And since customer_id becomes required with fieldservice_account_analytic, the test doesn't go further. I propose here to solve this by removing the dependency on fieldservice_account for fieldservice_account_analytic tests.


.travis.yml
```diff
@@ -24,6 +23,9 @@ jobs:
     - stage: linting
       name: "pre-commit"
       before_install:
+        - sudo add-apt-repository -y "deb https://apt-archive.postgresql.org/pub/repos/apt trusty-pgdg-archive main"
```
baimont: about this : postgis doesn't seem to be installable on trusty anymore.
https://www.postgresql.org/message-id/YBMtd6nRuXyU2zS4%40msg.df7cb.de

sbidoul:
Hi @moylop260, my colleage @baimont  is are trying to make this repo work on runbot.

The [problem](https://runbot.odoo-community.org/runbot/static/build/3481372-773-b5f8db/logs/job_10_test_base.txt) is that the vauxoo/odoo-80-image-shippable-auto that is used on runbot is based on Ubuntu 14 trusty and postgis is not easily available for that old OS anymore. Do you know if we can use another docker image on runbot, ideally based on an ubuntu version that is available on travis too ?

moylop260:
HI @sbidoul 

We need to work to use a new image

But I'm not able to do it right now

This is the project related with that image:
 - https://github.com/Vauxoo/docker-odoo-image/tree/master/odoo-shippable

It needs multi-python version in order to support to choose one

Maybe using a newer image and using [pyenv](https://github.com/Vauxoo/docker-odoo-image/blob/master/odoo110_py37/Dockerfile) could be a good approach

sbidoul:
Hi @moylop260 thanks for the explanation. It is probably not worth putting effort into this now, as we still plan to move to GitHub actions and the new runbot.

PR #782, opened 2021-05-06
Update fr.po

nexterite:
Hell
fr.po was rather empty, so I tried to fill in as much as I could. Some lines are still empty because I wasn't sure of the meaning in the context.


brian10048:
Hello and thank you for this work @nexterite 

It is my understanding that the translations are to be done on the OCA Weblate platform and will then be automatically merged to this repo
https://translation.odoo-community.org/projects/field-service-12-0/

PR #789, opened 2021-05-19
[14.0][FIX]fieldservice_timeline

Freni-OSI:
The issue has been solved in fieldservice (Ref #793 ), so no need to add in this module


smangukiya (approved):

smangukiya:
@Freni-OSI : Can you confirm this change is valid? I am not able to upgrade the fieldservice_timeline module after including this change.

CC: @murtuzasaleh

AmmarOfficewalaSerpentCS:
@smangukiya , 
yes, these valid changes,
 we move this code on #793 PR

Issue #802, opened 2021-06-04
[12.0] Gantt view (fsm_gantt.js) incompatible with date_range: error whenever field is empty


HekkiMelody:
The Gantt view found at Fieldservice > Dashboard > Orders, i.e. https://github.com/OCA/field-service/blob/12.0/fieldservice/static/src/js/fsm_gantt.js of the base fieldservice module, gives an error on page load, and every time the filter is set to empty afterwards, if date_range is also installed.

The error is triggered by the onchange method associated with the `<select>` element. You can find the element in question with 
`$(''#user_filer .o_searchview_extended_prop_field')`

If the onchange is called while the field is empty, this function from the date_range plugin is called with an empty "field" parameter, causing the error

https://github.com/OCA/server-ux/blob/af1984d5103c2ff148efd03616159fbcdc9e5777/date_range/static/src/js/date_range.js#L13-L22

The date_range module is a dependency of several other OCA modules.

I would be willing to work on a solution myself, but I'm not sure what it should look like or if it should be suggested here rather than in the date_range module.

A quick and dirty fix seems to be changing the date_range function like this:

```js
select_field: function (field) {
            this._super.apply(this, arguments);
            if (typeof field === 'undefined')
                return;
            this.is_date_range_selected = false;
            this.is_date = field.type === 'date' || field.type === 'datetime';
            this.$value = this.$el.find('.searchview_extended_prop_value, .o_searchview_extended_prop_value');
            if (this.is_date) {
                var ds = new data.DataSetSearch(this, 'date.range.type', this.context, [[1, '=', 1]]);
                ds.read_slice(['name'], {}).done(this.proxy('add_date_range_types_operator'));
            }
        },

```

but I'm not sure if that's desirable.

PR #807, opened 2021-06-09
[FIX][12.0]Fix gantt view error with date_range

HekkiMelody:
Fixed client error on fsm.order's Gantt View when date_range is also installed.

The issue was that the <select> element used as a filter could be empty. The onchange triggered when it was empty would cause the error. 
See also: #802

I considered proposing a fix on date_range, but this seems to be the only view with such an issue. These changes are also more consistent with the rest of the interface, where the <select> in question cannot be empty.


smangukiya (approved):
LGTM

PR #808, opened 2021-06-18
[13.0][FIX] fieldservice - in order to use request_early from vals, check request_early is there

baimont:
The check for vals.get("request_early") is only done when priority is 0 but should be done for all the cases.
Otherwise when you add timedelta you'll get : TypeError: unsupported operand type(s) for +: 'NoneType' and 'datetime.timedelta'


hparfr:
Can you please refactor with _compute_request_late ?

Btw, it's 8 hours not 8 days.

PR #762, opened 2021-04-15
[14.0][MIG] fieldservice_recurring

Freni-OSI:

PR #809, opened 2021-06-22
[14.0][add] fieldservice_contract

hparfr:
Blocked by fieldservice_recurring <s>#762</s> https://github.com/OCA/field-service/pull/859
Blocked by fieldservice_quick_edit not published in pr yet: https://github.com/akretion/field-service/tree/14.0-mig-fieldservice_recurring_quick_edit


fieldservice_contract/models/contract_line.py
```diff
@@ -0,0 +1,251 @@
...
+            result[line.id] = fsm_recurring
+        return result
+
+    def _field_service_generation(self):
+        """For service lines, create the field service order. If it already
+        exists, it simply links the existing one to the line.
+        """
+        for line in self:
+            # create order
+            if line.product_id.field_service_tracking == "order":
```
mourad-ehm: Field_service_tracking has not "order" value https://github.com/OCA/field-service/blob/14.0/fieldservice_sale/models/product_template.py#L17.  We have "sale" and "line" value which mean that we can create an unique FSM order by sale order if Field_service_tracking = "sale" or one SFM Order by sale line if Field_service_tracking = "line".
It seam that for recurring order we can only create one FSM recurring order by sale line.

fieldservice_contract/__manifest__.py
```diff
@@ -0,0 +1,33 @@
...
+    "license": "AGPL-3",
+    "version": "14.0.1.0.0",
+    "depends": [
+        "contract",
+        "product_contract",
+        "fieldservice_sale",
+        "fieldservice_recurring",
+        "fieldservice_sale_recurring",
+        "fieldservice_sale_frequency",
+        "fieldservice_recurring_quick_edit",
```
petrus-v, line 19: I could not find any PR to this module fieldservice_recurring_quick_edit on version 14.0, does this dependency is required ? or could you point give me the link that I could give a review?

fieldservice_contract/models/sale_order_line.py
```diff
@@ -0,0 +1,60 @@
...
+            "Realised fsm order not smoothing bill: This means that the amount of invoice don't depends on the number of fsm order Realised in the invoiced period.\n",
+            #"Realised fsm order smoothing bill: This means that the amount of invoice depends on the number of fsm order Realised in the invoiced period.\n",
+        default="order_smoothing_bill",
+    )
+
+    def _prepare_contract_line_values(
+        self, contract, predecessor_contract_line_id=False
+    ):
+        self.ensure_one()
+        contract.fsm_location_id = self.order_id.fsm_location_id
```
petrus-v, line 44: this is a little trap, as developer I won't expect the contract to be updated while preparing conctract liste, also it make hard to overload in submodules

fieldservice_contract/__manifest__.py
hparfr: https://github.com/akretion/field-service/tree/14.0-mig-fieldservice_recurring_quick_edit

hparfr: The dependency tree should be clarified / refactored.

First, we need `frequencies set` probably in his own module. Then defined in `sale order lines`, `contracts lines`.

`Frequency set` has two notions; 
- 1) abstract frequency set ( ambigus defined during the sale. Like "once a month".
- 2) concret frequency set, usually derived from the abstract. Like "first monday of each month at 6am".

fsm_recurring_quick_edit let you input an abstract frequency set` on the `sale order line` and convert it to a `concret frequency set` on a `recurring order`.

The abstract frequency set may be also used on the contract line.

PR #815, opened 2021-06-28
[12.0][FIX]Fixed _compute_request_late methods

HekkiMelody:
- There's a `_compute_request_late` method in the fieldservice/models/fsm_order.py model, but it's not used. In the `create` the same code is duplicated. This PR avoids duplication and also allows additional modules to easily access and modify the method itself, since it's not embedded in the `create` method anymore.

- The `_compute_request_late` method doesn't always check that `request_early` exists. Now it's checked at the start See also #808 for a similar fix on version 13.0

- Related fix in fieldservice_recurring/models/fsm_order.py : this module extends the method, but the signature doesn't match that of the original method, causing errors. The way the method was structured also didn't reflect how the original method works, so it has been updated.

### DIFF
--- a/fieldservice_recurring/models/fsm_order.py
+++ b/fieldservice_recurring/models/fsm_order.py
@@ -12,15 +12,19 @@ class FSMOrder(models.Model):
     fsm_recurring_id = fields.Many2one(
         'fsm.recurring', 'Recurring Order', readonly=True)
 
-    def _compute_request_late(self):
-        for rec in self:
-            if not rec.fsm_recurring_id:
-                return super(FSMOrder, self)._compute_request_late()
-            else:
-                days_late = rec.fsm_recurring_id.fsm_frequency_set_id.\
-                    buffer_late
-                rec.request_late = rec.scheduled_date_start + timedelta(
-                    days=days_late)
+    def _compute_request_late(self, vals):
+        if not vals.get("fsm_recurring_id", False):
+            return super(FSMOrder, self)._compute_request_late(vals)
+        elif vals.get("scheduled_date_start", False):
+            days_late = (
+                self.env["fsm.recurring"]
+                .browse(vals["fsm_recurring_id"])
+                .fsm_frequency_set_id.buffer_late
+            )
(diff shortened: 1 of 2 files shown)


HekkiMelody:
As an additional note, I'm not sure why the method in the fieldservice_recurring module uses `scheduled_date_start` as the starting point, rather than the `request_early field`. Let me know if you think I should change it...

hparfr:
> As an additional note, I'm not sure why the method in the fieldservice_recurring module uses `scheduled_date_start` as the starting point, rather than the `request_early field`. Let me know if you think I should change it...

When generated by recurring orders, you may want to keep the scheduled_date_start because it's already agreed with the customer.

In my setup I do not use request_early because the I find it too simple / too limited. I use the stage to know if the scheduled_date is confirmed or not.

hparfr (approved):

fieldservice_recurring/models/fsm_order.py
```diff
@@ -12,15 +12,19 @@ class FSMOrder(models.Model):
...
-                    buffer_late
-                rec.request_late = rec.scheduled_date_start + timedelta(
-                    days=days_late)
+    def _compute_request_late(self, vals):
+        if not vals.get("fsm_recurring_id", False):
+            return super(FSMOrder, self)._compute_request_late(vals)
+        elif vals.get("scheduled_date_start", False):
+            days_late = (
+                self.env["fsm.recurring"]
+                .browse(vals["fsm_recurring_id"])
```
hparfr, line 21: why a browse() here ?

HekkiMelody:
Sorry for the late reply. WIll rebase and address the comments asap.

fieldservice_recurring/models/fsm_order.py
HekkiMelody: The method is called from the create method, at this point self has no values on it. I'm forced to "browse" to find the fsm.recurring object, self.fsm_recurring_id is always empty.

HekkiMelody:
@dreispt I've rebased on 12.0 and pushed, there were no conflicts so I made no changes myself.

PR #822, opened 2021-07-13
[14.0][FIX] fieldservice: Purchase setting sets non-existant group

mtelahun:
Enabling the Purchase setting installs the purchase module and
tries to enable group_manage_vendor_price. However, this
group name was changed several years ago. It should set
group_product_pricelist instead.


mtelahun:
Sorry, tested this in a polluted environment, which caused the test to pass. Also, upon further investigation the implementation is wrong.

PR #826, opened 2021-07-17
[14.0][FIX][fieldservice] Add missing module dependency on OCA/stock_…

robyf70:
…request stock-logistics-warehouse

Fix missing ``stock_request`` dependency since ``res.config.settings`` uses ``group_stock_request_order`` 
https://github.com/OCA/field-service/blob/8260d64c4daaf32818a50250ba487724216c6f5b/fieldservice/models/res_config_settings.py#L112
that is declared at https://github.com/OCA/stock-logistics-warehouse/blob/89c50447b443fc426ef7f7f72cbfc8e58787e346/stock_request/models/res_config_settings.py#L10

### DIFF
--- a/fieldservice/__manifest__.py
+++ b/fieldservice/__manifest__.py
@@ -8,7 +8,7 @@
     "category": "Field Service",
     "author": "Open Source Integrators, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/field-service",
-    "depends": ["base_territory", "base_geolocalize", "resource", "contacts"],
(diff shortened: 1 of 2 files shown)


pedrobaeza:
The main module shouldn't depend on such module. It should be a glue module and extract the dependent code to it.

mtelahun:
I think that Instead of trying to set the group you should change the setting to install the module.

pedrobaeza:
Fixed in dc9c06b

PR #837, opened 2021-09-03
[14.0][IMP] base_territory:- Portal user access

murtuzasaleh:
Added Missing Portal user access right

https://github.com/OCA/field-service/blob/12.0/fieldservice/security/ir.model.access.csv#L19

https://github.com/OCA/field-service/blob/12.0/fieldservice/security/ir.model.access.csv#L22

https://github.com/OCA/field-service/blob/12.0/fieldservice/security/ir.model.access.csv#L25

https://github.com/OCA/field-service/blob/12.0/fieldservice/security/ir.model.access.csv#L28


smangukiya:
Managing portal access differently!

PR #839, opened 2021-09-08
[13.0][ADD] fieldservice - Added OU Script

Chanakya-OSI:
Added missing openupgrade script for renamed fields and models.


smangukiya:
@Chanakya-OSI Can you please check Travis CI and fix it so we can merge it? Thanks

PR #851, opened 2021-10-05
[12.0][FIX] Removed forced overwrite of fsm.order's Scheduled Date End

HekkiMelody:
Currently during FSM.Order's creation, the scheduled date end is forcibly overwritten by this call.
In addition, I cannot find any module that sets the "default_scheduled_date_end" key in the context that these lines refer to.

Not seeing the point of these lines of code anymore, I suggest removing them.


hparfr (approved):

marcelsavegnago (approved):
LGTM

PR #852, opened 2021-10-05
[12.0][IMP] Removed automatic Scheduled Date Start from FSM.Order Creation

HekkiMelody:
Reasoning for this change:

- it is a bad default. If we accept all the default values, say I'm creating an order at 10.30am, both the request early and the scheduled_date_start are set to 10:00am. You'd be scheduling something for the past.
- if request_early should be stripped of minutes and seconds, it should be done when setting the default value, and not on the create method of the order, when it could be overwriting something manually set by the user
- Scheduling should be something that is done manually by a user, and not set automatically, even when the user unsets the field on purpose.
- Maybe a better solution could be to add a setting along the lines of "automatically schedule for the earliest date" and optionally activate this mechanism again. I would be willing to develop that if it's something you think would improve the module.

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -89,7 +89,8 @@ class FSMOrder(models.Model):
                                   index=True, required=True)
     location_directions = fields.Char(string='Location Directions')
     request_early = fields.Datetime(string='Earliest Request Date',
-                                    default=fields.Datetime.now)
+                                    default=lambda _: fields.Datetime.now().
+                                    replace(second=0))
     color = fields.Integer('Color Index')
     company_id = fields.Many2one(
         'res.company', string='Company', required=True, index=True,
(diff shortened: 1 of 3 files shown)


hparfr (approved):

HekkiMelody:
@dreispt Rebased and solved conflicts

dreispt:
Thank you. Travis is still not happy, though.

HekkiMelody:
Ah, tests need to be updated. Will get on it asap.

HekkiMelody:
Should be better now. Afaict, the failures are not due to my changes...? Do tell if I'm mistaken, though.

hparfr:
@aleuffre the error is : 

```
 ======== Testing test_flake8 ========

fieldservice/models/fsm_order.py:92:89: E501 line too long (94 > 88 characters)
```

HekkiMelody:
Oof, thank you. @hparfr :pray:

HekkiMelody:
Travis seems happy, finally :tada: .
Thank you all for the help and pointers.

PR #861, opened 2021-12-08
Correcting acentuation portugues to be redable

marcos-mendez:
This is trying to prevent encoding error while portugues version is not finished to make the fields readable by the user

Screenshot from 2021-12-08 16-19-35
Any other acentuation problems


marcelsavegnago:
@marcos-mendez Translations and corrections of translations must be done via weblate and not via PR.

Please see it https://odoo-community.org/page/translate

PR #864, opened 2021-12-13
[14.0][FIX] Fixed an issue of type error while on form view load

Vimal-OSI:


fieldservice_stock/views/fsm_order.xml
```diff
@@ -49,7 +49,7 @@
                     >
                         <tree
                             decoration-muted="state == 'cancel'"
-                            default_order='date_expected, picking_id, sequence'
+                            default_order='date, picking_id, sequence'
```
chienandalu, line 52: Shouldn't it be `date_deadline`? (https://github.com/OCA/OpenUpgrade/blob/0fea456fb0ad2b9a4cc13798f9fbdd474d132132/openupgrade_scripts/scripts/stock/14.0.1.1/pre-migration.py#L14)

Vimal-OSI: Yeah,correct! It should be `date_deadline` fields, But in tree view isn't  `date_deadline` field and used the `date` field is in tree view, so that, I was thought that better to add date field in place of deadline.( I think It gives an error if we will use "date_deadline" without add on tree view)

chienandalu: Sure it will throw an error. Thanks for the explanation :+1:

chienandalu (approved):

PR #875, opened 2022-01-04
[14.0][FIX] fieldservice_sale

murtuzasaleh:
We will find out the solution to Fix the Issue, currently, I have commented on this line because it is causing the Issue in QA and Production Database.

odoo.models.unlink: User #1 deleted ir.config_parameter records with IDs: [25] 
odoo.modules.loading: Module fieldservice_sale loaded in 1.06s, 939 queries 
odoo.addons.base.models.ir_module: ALLOW access to module.module_uninstall on ['Field Service Web Timeline'] to user __system__ #1 via n/a

### DIFF
--- a/fieldservice_sale/views/res_config_settings.xml
+++ b/fieldservice_sale/views/res_config_settings.xml
@@ -1,11 +1,13 @@
 <odoo>
 
-    <record model="res.config.settings" id="res_config_settings_fsm_template">
+    <!-- ToDo -->
+
+    <!-- <record model="res.config.settings" id="res_config_settings_fsm_template">
         <field name="group_fsm_template" eval="1" />
     </record>
 
     <function model="res.config.settings" name="execute">
             <value eval="[ref('fieldservice_sale.res_config_settings_fsm_template')]" />
-    </function>
+    </function> -->
 
 </odoo>
--- a/fieldservice_sale/__manifest__.py
+++ b/fieldservice_sale/__manifest__.py
@@ -18,7 +18,7 @@
         "views/fsm_order.xml",
         "views/product_template.xml",
         "views/sale_order.xml",
-        "views/res_config_settings.xml",
+        # "views/res_config_settings.xml",
     ],
     "license": "AGPL-3",
     "development_status": "Beta",


murtuzasaleh:
@dreispt @smangukiya Can you please review this PR?

fieldservice_sale/views/res_config_settings.xml
```diff
@@ -1,11 +1,13 @@
...
 
-    <record model="res.config.settings" id="res_config_settings_fsm_template">
+    <!-- ToDo -->
+
+    <!-- <record model="res.config.settings" id="res_config_settings_fsm_template">
         <field name="group_fsm_template" eval="1" />
     </record>
 
     <function model="res.config.settings" name="execute">
             <value eval="[ref('fieldservice_sale.res_config_settings_fsm_template')]" />
```
dreispt, line 10: You can try using set_values instead.
Note: both `execute` and `set_values` don't have any arguments...
```suggestion
    <function model="res.config.settings" name="set values">
```

dreispt (approved):
This is a workaround until a working solution is found to enable the FSM Templates by default.
Odoo Sales app does something similar with this code: https://github.com/odoo/odoo/blob/14.0/addons/sale_management/data/sale_order_template_demo.xml#L4

dreispt:
@murtuzasaleh Buiding is failing because of a missing dependency `agreement_serviceprofile`:
https://app.travis-ci.com/github/OCA/field-service/jobs/554177705#L555

dreispt:
Update: this is an issue with the OCA Wheelhouse when accessing  https://wheelhouse.odoo-community.org/oca-simple-and-pypi/odoo14-addon-agreement-serviceprofile.
I reported it.

hparfr:
Hi,

Can you give us more info about how to re-produce the bug ?

hparfr:
Hey, why merging now ?
I did not got an answer, and the PR is adding dead code.

smangukiya:
@hparfr It was a workaround and until we fix the FSM template as Daniel said earlier. If you still thinks that it will not work for you then feel free to revert it back and submit a PR we will revert this back. Thank you

"This is a workaround until a working solution is found to enable the FSM Templates by default."

hparfr:
@murtuzasaleh  @smangukiya @dreispt does have one of you the opportunity to fix the regression introduced in this PR  ?

```
************* Module fieldservice_sale
fieldservice_sale/views/res_config_settings.xml:1: [W7930(file-not-used), ]  not used from manifest
```

 https://github.com/OCA/field-service/runs/5760987793?check_suite_focus=true#step:7:103

It's blocking https://github.com/OCA/field-service/pull/923

PR #890, opened 2022-01-24
[14.0][FIX] fieldservice: Apply sudo if user has not access to fsm.location model.

victoralmau:
Apply `sudo()` if user has not access to `fsm.location` model.

Please @CarlosRoca13 and @pedrobaeza can you review it?

@Tecnativa TT34048

### DIFF
--- a/fieldservice/models/res_partner.py
+++ b/fieldservice/models/res_partner.py
@@ -59,6 +59,7 @@ class ResPartner(models.Model):
         wiz = self.env["fsm.wizard"]
         partners_with_loc_ids = (
             self.env["fsm.location"]
+            .sudo()
             .search([("active", "in", [False, True]), ("partner_id", "in", self.ids)])
             .mapped("partner_id")
         ).ids


pedrobaeza:
Are you sure this is the solution instead of limiting by groups in the view the available smart-buttons?

victoralmau:
> Are you sure this is the solution instead of limiting by groups in the view the available smart-buttons?

`_convert_fsm_location()` function is called in `write()` function from partners. IMO there is no other option for this.

fieldservice/models/res_partner.py
```diff
@@ -57,10 +57,13 @@ def action_open_owned_locations(self):
 
     def _convert_fsm_location(self):
         wiz = self.env["fsm.wizard"]
+        location_model = self.env["fsm.location"]
+        if not self.env.user.has_group("fieldservice.group_fsm_user_own"):
```
pedrobaeza: No need to do this. Put always `sudo`, as since 13+, no user spoofing is done. Only the permission part.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #909, opened 2022-02-14
[14.0][FIX] fieldservice: loading issue while add multiple sub-location

Vimal-OSI:
Reproduce Steps:
 - Create Main FSM Location
 - Create 200+ Sublocation of Main location
 - Open the Main location record

### DIFF
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -10,6 +10,11 @@ class FSMLocation(models.Model):
     _inherits = {"res.partner": "partner_id"}
     _inherit = ["mail.thread", "mail.activity.mixin"]
     _description = "Field Service Location"
+    _parent_name = "fsm_parent_id"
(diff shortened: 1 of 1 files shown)


osi-scampbell (approved):

osi-scampbell:
@Vimal-OSI I believe this code only goes 1 level down correct? As in, if I have Location A and it is a parent of Location B, and Location B is a parent of Location C. This code will only show Location A has 1 sublocation, when in fact, it has 2. Does that make sense?

Vimal-OSI:
> @Vimal-OSI I believe this code only goes 1 level down correct? As in, if I have Location A and it is a parent of Location B, and Location B is a parent of Location C. This code will only show Location A has 1 sublocation, when in fact, it has 2. Does that make sense?

This code goes N level down, Location A shows Location B and C as sublocation. (Location A shows 2 sublocation).

PR #910, opened 2022-02-15
[14.0][IMP] fieldservice_account: Added search attribute to filter out orders on invoices

Vimal-OSI:


brian10048:
I do not get the expected result when I apply a filter such as "is not set" or "is not equal to" 

The FSM order is still in the list when it should be excluded. 

## No filter
account search fsm

## Filter = FSM is set
**Expected Result**
account search fsm set

## Filter = FSM is not set
**Unexpected result.  The invoice with an FSM order should not be in the list**
account search fsm not set 

## Filter = FSM is equal to
**Expected Result**
account search fsm equal

## Filter = FSM is not equal to
**Unexpected result.  The invoice with the same FSM Name as the filter should not be in the list**
account search fsm not equal

PR #914, opened 2022-03-10
[13.0][FIX] fieldservice_sale - confirm quotation with section or note lines

baimont:
When you use a product that should not create a fsm order and add a section line or a note line in a quotation, odoo raises an error asking for a FSM Location when you try to confirm the quotation. But you should be able to confirm the sale order without FSM Location since there will not be any field service creation.

fieldservicenoproduct
saleorderno


apineux (approved):

baimont:
runboat fails : https://github.com/sbidoul/runboat/issues/48

EDIT: fixed

rousseldenis (commented):
@baimont Could be a quick win to add a test with sections

hparfr (approved):

PR #919, opened 2022-03-15
[14.0] [FIX] fieldservice_recurring: Attempt to fix daylight saving time jumps

paradoxxxzero:
Use timezone aware rrule and normalize recurring dates to avoid hour jumps during daylight saving time changes.

Without this PR when recurring weekly between 2022-03-21 15:00:00 and 2022-04-11 15:30:00 in timezone Europe/Paris we get:
```
2022-03-21 15:00:00
2022-03-28 16:00:00
```
with it we get:
```
2022-03-21 15:00:00
2022-03-28 15:00:00
```

### DIFF
--- a/fieldservice_recurring/models/fsm_frequency.py
+++ b/fieldservice_recurring/models/fsm_frequency.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2019 Brian McMaster, Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import pytz
 from dateutil.rrule import (
     DAILY,
     FR,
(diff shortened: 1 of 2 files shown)


smangukiya (approved):
LGTM

smangukiya:
@brian10048 if you can review this too. Thanks

hparfr (approved):

hparfr:
Tests needs to be fixed.

` fields.Datetime.today() ` always start at 0:0, so I guess if we change the start hour of the tests there should be no problem.

smangukiya:
@paradoxxxzero Thank you for your work here, can you please attend @hparfr 's comment?

brian10048 (approved):

PR #882, opened 2022-01-07
[15.0][MIG] fieldservice_vehicle

Vandan-OSI:

PR #821, opened 2021-07-06
[14.0][MIG] fieldservice_vehicle_stock

ReedHayashikawa:
Depends on: 
[OCA/stock-logistics-warehouse#1193](https://github.com/OCA/stock-logistics-warehouse/pull/1193)
#730 
#690

PR #921, opened 2022-03-25
[WIP][MIG][15.0] fieldservice_vehicle_stock

Vandan-OSI:


Vandan-OSI:
@max3903  @dreispt  @brian10048 

This PR depends on:
- [x] https://github.com/OCA/field-service/pull/882
- [x] https://github.com/OCA/stock-logistics-workflow/pull/1095

brian10048:
@max3903 
I'm trying to understand how the putaway features of this module function and if the dependency on `stock_putaway_method` is still required.  

Core Odoo removed the method `putaway_apply` after v12 and this method was where the putaway method added by `stock_putaway_method` gets used.  During migration of this module from v12 to v14 ([here](https://github.com/OCA/field-service/pull/821/commits/c28f2ae8768d6ac4647eb48746fd57f2e68172b8#diff-428c72e64e9870008b21ec1160b74cfa9db8490913e3b70421627ea28392290d)) the method `putaway_apply` was deleted since it's no longer in core Odoo to override but the putaway feature was never replaced.

I think this module should be refactored to remove the dependency (see #945) or the dependency could be improved to once again provide the functionality `putaway_apply` had when it was a part of core Odoo

Vandan-OSI:
@max3903  As per @brian10048  Feedback Do We Still Need a stock_putaway_method module Dependency?

max3903:
@Vandan-OSI @brian10048 The `stock_putaway_method` module is used to determine the location of the specific vehicle when the location is set to the generic "Vehicles" location. We should be using `stock_putaway_hook` https://github.com/OCA/stock-logistics-workflow/tree/15.0/stock_putaway_hook to provide the same feature.

PR #924, opened 2022-04-07
[FIX][14.0] fielservice_agreement fix pre-commit

hparfr:
It will be better to properly fix by not using replace I guess

Resolves parts of #923


smangukiya (approved):
LGTM

brian10048 (approved):

PR #925, opened 2022-04-08
[14][FIX] fieldservice_recurring: fsm_order_ids should be readonly

hparfr:
in the UI, it should not be allowed to select orders to be added to recurring.


brian10048:
@hparfr I don't intend to add a new order to a recurring in the UI, but it is helpful to be able to edit an existing order from the view of the recurring.  For example to edit the assigned worker.

hparfr:
Fixed, the widget was not the good one, so create=false was not effective.

I allowed to unlink() from recurring.

brian10048 (approved):
Tested on runboat

PR #940, opened 2022-06-07
[14.0][REF] fieldservice_sale: Refactor FSM Order generation from Sales

brian10048:
This PR aims to:

- Remove similar code in sale order and sale order line.  Field service generation is now done by the sale order

- Improve relationships between sale order, sale order line, and fsm order.  Now a field service order will be linked to every sale order line that generates a field service order, even in the case of field service tracking = sale, which were previously only linked to the sale and not any of the lines

The main reason for this change is to simplify the field service generation code that was similar between sale order and sale order line.  

The only change in functionality is for products with fsm_tracking = sale.
Previously, only one FSM Order per Sale Order was created regardless of the FSM Template
Now, one FSM Order for each unique Template per Sale Order will be created.


hparfr:
Interesting and thanks to address this subject.

I don't see the use case for the relation Multiple SO <-> One FSM Order 

In my uses cases, I don't use the setting "create one FSM Order per sale order". It blocks the ability to invoice the FSM Order. So if we agree to get rid of it can also simplify the code and understanding of FSM for new comers.

I use the module  `fsm_sale_frequency` (not yet published to oca because it needs some refactoring). It adds one really interesting feature : the ability to define a frequency set on the sale order line : 
https://github.com/akretion/field-service/blob/14_add_fieldservice_sale_frequency/fieldservice_sale_frequency/models/sale_order_line.py#L11 

In products I have always the same policy: if there is a frequency on the sol, create a recurring order otherwise create a fsm order  https://github.com/akretion/field-service/blob/14_add_fieldservice_sale_frequency/fieldservice_sale_frequency/models/product_template.py#L13

For example, 
Plant a tree -> One FSM Order
Prune the branches (Twice a year) -> One recurring -> 2 FSM Orders
image

Other example,
Plant a tree -> One FSM Order
Prune the branches (once) -> One FSM Orders
image

Because the same product (prune the branches) can be done once or multiple time (recurring). It's the same product, it's a setting on the sol.

With this setup, we can simplify the tracking policy to only:

fsm_sale installed:
* don't track FSM
* track FSM on SOL

fsm_sale + fsm_sale_recurring installed:
* don't track FSM
* track FSM on SOL *(create the recurring if there is a frequency set on the sol).*


What do you think @brian10048 ?

brian10048:
> I don't see the use case for the relation Multiple SO <-> One FSM Order

I think that having multiple SO to one FSM order allows for the greatest flexibility.  The need for this arose from this specific workflow:

1. Customer confirms a "One Time Maintenance" service that costs a flat fee and a FSM Order is generated from the SO
2. Worker performs the maintenance service and at that time confirms with customer some additional parts are needed at additional cost.
3. From the FSM Order, a second SO is created for these added parts.

Both SO are linked to the same FSM Order that represents when the work was completed.


>In my uses cases, I don't use the setting "create one FSM Order per sale order". It blocks the ability to invoice the FSM Order. So if we agree to get rid of it can also simplify the code and understanding of FSM for new comers.

I use the one FSM Order per SO setting on several products so I'm not quite ready to agree to get rid of it.  This PR gets rid of a bit of similar code so that should be quite helpful as a start.

hparfr:
> Worker performs the maintenance service and at that time confirms with customer some additional parts are needed at additional cost.
> From the FSM Order, a second SO is created for these added parts.

If the worker returns another day when he has the additional parts, it not the same fsm_order.
Your problem is how to generate new parts delivery or (new fsm_order) from an existing fsm order and may be invoice this new delivery.

[Does stock_picking_invoicing](https://github.com/OCA/account-invoicing/tree/14.0/stock_picking_invoicing) solves it ? 


>Both SO are linked to the same FSM Order that represents when the work was completed.

It's not the same kind of link.
- link we have: SOL -generate> FSM.order 
- link you describe: FSM.order(s) -reference> SO


>I use the one FSM Order per SO setting on several products so I'm not quite ready to agree to get rid of it.

Ok

brian10048:
superseded by #1049 without the relationship changes

PR #943, opened 2022-06-21
[FIX][14.0]: fieldservice_sale_recurring: recurring order validation …

mourad-ehm:
recurring order validation generate wrong fsm_order (ex: Assigned To not set, start date wrong, recurring info not set, ...)


hparfr (approved):

brian10048:
Can we make this a setting that lets each company decide if they would like to start recurring orders upon sale confirmation or not immediately start them?

hparfr:
In the recurring order there is some fields to complete: assigned worker, adjust the frequency, start and end date, max orders, ... 

@brian10048   Do you have some use cases where it's wanted to start them immediately ? Is it worth the complexity in a generic module ?

hparfr:
@brian10048 Do you have some use cases where it's wanted to start them immediately ? Is it worth the complexity in a generic module ?

brian10048:
No, I think this will be fine

brian10048 (approved):

PR #945, opened 2022-06-24
[14.0] [IMP] fieldservice_vehicle_stock

brian10048:
This module had limited functionality previously and the product putaway feature became useless when migrated to v14.

This PR removes dead code and the extra dependency.

Stock operation types can now be configured for FSM Vehicle loading operations rather than being hardcoded to the records added by this module.

New functionality has been added to help automate the storage location of inventory moves destined for FSM Vehicles.

The README has been expanded to further clarify the purpose of this module and tests were completely rewritten


hparfr (approved):
Code review

BTW, why putaway is not needed anymore ?

brian10048:
@hparfr I removed putaway because it no longer functions as it did in v12.

This module and the dependency `stock_putaway_method` were both migrated directly to 14 from 12.  The migration of `stock_putaway_method` [here](https://github.com/OCA/stock-logistics-warehouse/commit/ce255ebdbad866dcfa90af08094a2aba2d26f6bf) only changes the model name and doesn't account for the functional changes done in core Odoo between version 12 and 14 and therefore it doesn't serve any function that I am aware of.

[Odoo Product Strategy v12](https://github.com/odoo/odoo/blob/12.0/addons/stock/models/product_strategy.py)
[Odoo Product Strategy v14](https://github.com/odoo/odoo/blob/14.0/addons/stock/models/product_strategy.py)

max3903:
@brian10048 It looks like the destination location is set when the picking is done. How do you know which vehicle needs to be loaded? The original feature is that the vehicle/location is set when the picking is reserved.

brian10048:
> @brian10048 It looks like the destination location is set when the picking is done. How do you know which vehicle needs to be loaded? The original feature is that the vehicle/location is set when the picking is reserved.

@max3903 Are you referring to the change I made [here?](https://github.com/OCA/field-service/pull/945/files#diff-6c747e63dd0f0add38d49ae502bb7149899e29af32a63b82d8c642b89a7ca9efR12) All that was happening before this change was to validate that we had the FSM vehicle set for specific picking types.  I decided to change the validation to occur when the picking is done, because we often don't know which specific vehicle will be loaded when the product is initially reserved.  For example, I would always receive the User Error when confirming a sale order that attempts to reserve some product for a field service order.

I determine which vehicle to be loaded when a vehicle is assigned to the FSM Order associated with the picking.

Also check https://github.com/OCA/field-service/pull/921#issuecomment-1311865015

max3903:
@brian10048

> I determine which vehicle to be loaded when a vehicle is assigned to the FSM Order associated with the picking.

Can you exactly do that instead of setting it after the picking is done? I mean set the vehicle stock location on the stock move line when the vehicle is set on the FSM order. This way we can support both scenario.

PR #953, opened 2022-08-22
[ADD] fieldservice_stage_close_date

BojanOD:
Datetime field added to the fsm.order that tells the exact time when an order has been closed.
An order has been closed if the order stage is_closed.

### DIFF
--- a/fieldservice_fsm_stage/__init__.py
+++ b/fieldservice_fsm_stage/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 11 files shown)


damdam-s:
Can you please add unit tests on this?

brian10048:
@BojanOD thank you for this addition.  Everything looks good.

I would only suggest a more descriptive module name such as `fieldservice_stage_close_date` and squash commits once completed

marcos-mendez (approved):
LGTM, Tested under LXC and RB

dreispt (commented):
Why not just adding this directly to the `fieldservice` base module?

damdam-s:
reason is to have simple modules and not a big `fieldservice` one. 
if the PSC thinks it's better to incorporate this one to the main one, no issue.
We did the "split" for modularity reasons and not to have this behaviour mandatory

dreispt:
That is true, but OTOH a module should provide a reasonably feature set.
No strong opinion myself on this.

PR #957, opened 2022-09-06
[FIX] fieldservice: Duplicate/copy method

BojanOD:
Don't show

> Equipment name already exists!

when duplicating/copying an equipment.

### DIFF
--- a/fieldservice/models/fsm_equipment.py
+++ b/fieldservice/models/fsm_equipment.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2018 - TODAY, Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
 
 
 class FSMEquipment(models.Model):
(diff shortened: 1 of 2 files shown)


fieldservice/models/fsm_equipment.py
```diff
@@ -102,3 +102,9 @@ def _onchange_stage_id(self):
             self.hide = True
         else:
             self.hide = False
+
+    def copy(self, default=None):
+        if default is None:
+            default = {}
+        default["name"] = self.name + _(" (copy)")
+        return super(FSMEquipment, self).copy(default=default)
```
yankinmax:
```suggestion
        default = default if default else {}
        default["name"] = self.name + _(" (copy)")
        return super().copy(default=default)
```

fieldservice/models/fsm_equipment.py
```diff
@@ -102,3 +102,9 @@ def _onchange_stage_id(self):
             self.hide = True
         else:
             self.hide = False
+
+    def copy(self, default=None):
+        if default is None:
+            default = {}
+        default["name"] = self.name + _(" (copy)")
```
yankinmax: what does `" (copy)"` string bring?
what is the expected result of this concatenation?

yankinmax: ah ok, I see)
`EQ1 (copy)` shouldn't raise an exception

ivantodorovich (approved):

yankinmax (approved):

brian10048 (approved):

ivantodorovich (approved):

hparfr (commented):
/ocabot merge minor

PR #958, opened 2022-09-14
[FIX] base_territory: formatting

max3903:


base_territory/models/res_country.py
```diff
@@ -1,6 +1,5 @@
 # Copyright (C) 2020 Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
max3903, line 2:
```suggestion
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).

```

max3903:
@jantoniorivera97 Please review...

PR #964, opened 2022-10-06
[15.0][MIG] fieldservice_route: Migration to 15.0

victoralmau:
Migration to 15.0

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT35715


dreispt (approved):

fieldservice_route/models/fsm_order.py
```diff
@@ -14,11 +14,16 @@ class FSMOrder(models.Model):
...
+        comodel_name="fsm.route.dayroute", string="Day Route", index=True
+    )
     fsm_route_id = fields.Many2one(related="location_id.fsm_route_id", string="Route")
 
     person_id = fields.Many2one(
-        "fsm.person", string="Assigned To", index=True, default=_get_default_person
+        comodel_name="fsm.person",
+        string="Assigned To",
+        index=True,
+        default=_get_default_person,
```
pedrobaeza: Please use lambda

pedrobaeza (changes requested):
Please change all the default to lambdas

dreispt:
@pedrobaeza why would lambda usage be required?

pedrobaeza:
For having each time evaluated the value, instead of only on instance initialization. See also the guidelines for that: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#37models

fieldservice_route/models/fsm_order.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (C) 2019 Open Source Integrators
 # Copyright (C) 2019 Serpent consulting Services
+# Copyright 2022 Tecnativa - Víctor Martínez
```
nikul-serpentcs: @victoralmau Thank you for PR.

Can you please remove your **Copyright** and add in the contributor's fragment?

victoralmau: Changes done.

pedrobaeza: If he changes the file, the copyright should be there.

fieldservice_route/models/fsm_order.py
```diff
@@ -14,13 +14,23 @@ class FSMOrder(models.Model):
     def _get_default_person(self):
         return self.fsm_route_id.fsm_person_id.id or False
```
pedrobaeza: This is not going to work, as `fsm_route_id` won't never have a value on default. This must work with context values or previous defaults.

fieldservice_route/models/fsm_order.py
```diff
@@ -14,13 +14,23 @@ class FSMOrder(models.Model):
...
     person_id = fields.Many2one(
-        "fsm.person", string="Assigned To", index=True, default=_get_default_person
+        comodel_name="fsm.person",
+        string="Assigned To",
+        index=True,
+        default=lambda self: self._get_default_person(),
     )
 
+    @api.onchange("fsm_route_id")
+    def _onchange_person_id(self):
```
pedrobaeza: Can you do it a computed writable field instead?

pedrobaeza (commented):
Please squash the initial commits from Maxime that are the same into one.

victoralmau:
Changes done.

pedrobaeza:
The converted computed field is not computed writable, and I still see a lot of repeated commits from Maxime.

victoralmau:
> The converted computed field is not computed writable, and I still see a lot of repeated commits from Maxime.

Sorry but the `person_id` field is now compute `store=True`, right? (although in `fieldservice` addon it is many2one )

I have updated the commit history to reduce Maxime's commits.

pedrobaeza:
But it misses readonly=False

victoralmau:
> But it misses readonly=False

Ok, changes done.

CarlosRoca13 (approved):

pedrobaeza (approved):

PR #976, opened 2022-10-26
[IMP] fieldservice_stage_validation - validate if validate fields

baimont:
This PR updates black version and adds a condition in fieldservice_stage_validation

### DIFF
--- a/fieldservice_stage_validation/models/fsm_location.py
+++ b/fieldservice_stage_validation/models/fsm_location.py
@@ -13,6 +13,8 @@ class FSMLocation(models.Model):
             stage = rec.stage_id
             field_ids = stage.validate_field_ids
             field_names = [x.name for x in field_ids]
+            if not field_names:
+                continue
             values = rec.read(field_names)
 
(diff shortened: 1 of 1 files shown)


apineux (approved):

pedrobaeza:
You shouldn't change black version on your own, but stick to copier template updates, and please only one change per PR (or the fix, or the copier update, not both).

baimont:
> You shouldn't change black version on your own, but stick to copier template updates, and please only one change per PR (or the fix, or the copier update, not both).

you're right @pedrobaeza, I adapted this pr

fieldservice_stage_validation/models/fsm_location.py
```diff
@@ -13,6 +13,8 @@ def _validate_stage_fields(self):
             stage = rec.stage_id
```
pedrobaeza, line 13: Maybe a bit cleaner is `for rec in self.filtered(lambda x: x.stage_id.validate_field_ids):`

PR #978, opened 2022-10-27
[IMP][14.0] fieldservice_sale: add fsm location on sale line

mourad-ehm:


max3903:
@mourad-ehm Since Odoo only supports one analytic account on a sale order, we always settled down for one sale order per FSM location.

So I am not against having many locations on a sales order, but we have to be clear in the documentation that the revenue from the sales order will go to one analytic account, even if the locations are related to different analytic accounts.

fieldservice_sale/models/sale_order.py
```diff
@@ -138,10 +138,14 @@ def _field_find_fsm_order(self):
...
-            if not self.fsm_location_id:
+        fs_sale_lines = self.order_line.filtered(
+            lambda l: l.product_id.field_service_tracking != "no"
+        )
+        if fs_sale_lines:
+            if (
+                any(not sol.fsm_location_id for sol in fs_sale_lines)
+                and not self.fsm_location_id
+            ):
                 raise ValidationError(_("FSM Location must be set"))
```
hparfr: May we have a different error message when there is no FSM in one sale order line (with some hints like line number or product);

max3903:
@mourad-ehm Not sure what you did, but such a big change (323 files) will never get reviewed and merged.

mourad-ehm:
hi @max3903, it was a rebase problem.

hparfr:
@mourad-ehm can you please fix the conflicts?

mourad-ehm:
replaced by https://github.com/OCA/field-service/pull/1095

PR #979, opened 2022-10-27
[16.0][MIG] fieldservice_account

ghost:


max3903:
@gbrito-erpopen Same feedback as in #980

PR #980, opened 2022-10-27
[16.0][MIG] fieldservice_activity

ghost:


max3903:
@gbrito-erpopen You need to keep the history of the module from the previous branch. Here is how:
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0#technical-method-to-migrate-a-module-from-150-to-160-branch

PR #997, opened 2022-11-03
[IMP] update dotfiles

brian10048:


pedrobaeza:
Please don't mix the update with changes not strictly needed for passing CI.

brian10048:
@pedrobaeza I only did so because the test failed regarding Stable modules depending on Beta modules

pedrobaeza:
But the `order` to `record` variable change + the test are not that changes. Maybe cross-PR commits?

pedrobaeza (approved):

PR #998, opened 2022-11-03
[IMP] fieldservice_account: Tests

brian10048:
Tests should test something

### DIFF
--- a/fieldservice_account/__manifest__.py
+++ b/fieldservice_account/__manifest__.py
@@ -15,6 +15,6 @@
         "views/fsm_order.xml",
     ],
     "license": "AGPL-3",
-    "development_status": "Beta",
+    "development_status": "Production/Stable",
     "maintainers": ["osimallen", "brian10048", "bodedra"],
 }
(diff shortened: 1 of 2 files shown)


fieldservice_account/__manifest__.py
```diff
@@ -15,6 +15,6 @@
         "views/fsm_order.xml",
     ],
     "license": "AGPL-3",
-    "development_status": "Beta",
+    "development_status": "Production/Stable",
```
ghost, line 18: Any reason to move this to Production/Stable?

max3903: @gbrito-erpopen This module is only needing tests to move to `Production/Stable". Makes sense to me. :+1:

ghost: As long as it has tests, the module is ready for Production/Stable, correct? clear

brian10048:
Tests depend on #1004

fieldservice_account/__manifest__.py
max3903: Correct.

bodedra (approved):
LG 👍

max3903 (approved):

PR #1002, opened 2022-11-08
 [16.0][FIX] fieldservice_stock: ensure inventory_location is assigned

ghost:
When install the module, we want to ensure that the existing fsm.location
records have the default inventory_location_id assigned.

### DIFF
--- a/fieldservice_stock/__init__.py
+++ b/fieldservice_stock/__init__.py
@@ -2,3 +2,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import models
+from .hooks import _pre_init_hook
--- a/fieldservice_stock/__manifest__.py
+++ b/fieldservice_stock/__manifest__.py
@@ -20,6 +20,7 @@
         "views/stock.xml",
         "views/stock_picking.xml",
     ],
+    "pre_init_hook": "_pre_init_hook",
     "license": "AGPL-3",
     "development_status": "Beta",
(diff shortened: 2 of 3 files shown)


max3903:
@gbrito-erpopen Let's keep the field required in the database and put the default value on the existing data with a hook.

ghost:
I see what you mean, I will add an hook, it will assign the default value on the empty ones.

ghost:
@max3903 I don't think that is the solution. Maybe i'm not seeing correctly but:
 - When the fieldservice_stock is installed, it creates a constrain in the database on the field inventory_location_id.
 - When I run the tests on the fieldservice_account, it tries to create an fsm.location(line 28), but since we are not aware of the fieldservice_stock, we don't add the field inventory_location_id, odoo triggers this  `psycopg2.errors.NotNullViolation: null value in column "inventory_location_id" of relation "fsm_location" violates not-null constraint
`

Some previous Odoo versions the required=True on the field only affected the frontend, but a couple of Odoo versions a go, it creates a constraint in the database.

At the moment I still see that only enable the required in the fronted would be the best solution. There are more modules that trigger this error.

I implemented the pre_init hook, to fill the available fsm.locations in the default stock location

max3903:
@gbrito-erpopen When the `fsm.location` is created after the installation of `fieldservice_stock`, the default value for the `inventory_location_id` field should be used.

ghost:
@max3903 @brian10048 for review :)

max3903 (approved):

PR #1019, opened 2022-12-12
[14.0][IMP] Fieldservice Stage Server Action

BojanOD:
Improvements to the `fieldservice_stage_server_action` module.

### DIFF
--- a/fieldservice_stage_server_action/models/fsm_equipment.py
+++ b/fieldservice_stage_server_action/models/fsm_equipment.py
@@ -1,0 +1,37 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class FSMEquipment(models.Model):
+    _inherit = "fsm.equipment"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        for record in records:
+            action = record.stage_id.action_id
+            if action:
+                context = {
+                    "active_model": self._name,
+                    "active_ids": [record.id],
+                }
+                action.with_context(**context).run()
+        return records
+
+    def write(self, vals):
+        res = super().write(vals)
+        action = (
+            self.env["fsm.stage"].browse(vals["stage_id"]).action_id
+            if "stage_id" in vals
+            else None
+        )
+        if action:
+            for record in self:
+                context = {
+                    "active_model": record._name,
+                    "active_ids": record.ids,
+                }
+                action.with_context(**context).run()
+        return res
--- a/fieldservice_stage_server_action/models/fsm_location.py
+++ b/fieldservice_stage_server_action/models/fsm_location.py
@@ -1,0 +1,37 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class FSMLocation(models.Model):
+    _inherit = "fsm.location"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        for record in records:
+            action = record.stage_id.action_id
+            if action:
+                context = {
+                    "active_model": self._name,
+                    "active_ids": [record.id],
+                }
+                action.with_context(**context).run()
+        return records
+
+    def write(self, vals):
+        res = super().write(vals)
+        action = (
+            self.env["fsm.stage"].browse(vals["stage_id"]).action_id
+            if "stage_id" in vals
+            else None
+        )
+        if action:
+            for record in self:
+                context = {
+                    "active_model": record._name,
+                    "active_ids": record.ids,
+                }
+                action.with_context(**context).run()
+        return res
--- a/fieldservice_stage_server_action/models/fsm_order.py
+++ b/fieldservice_stage_server_action/models/fsm_order.py
@@ -1,0 +1,37 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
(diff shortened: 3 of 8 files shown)


fieldservice_stage_server_action/models/fsm_equipment.py
```diff
@@ -0,0 +1,31 @@
...
+
+from odoo import api, models
+
+
+class FSMEquipment(models.Model):
+    _inherit = "fsm.equipment"
+
+    @api.model
+    def create(self, vals):
+        res = super(FSMEquipment, self).create(vals)
```
yankinmax:
```suggestion
    @api.model_create_multi
    def create(self, vals_list):
        res = super().create(vals_list)
        for record in res:
            action = self.env["fsm.stage"].browse(record.stage_id).action_id
```

fieldservice_stage_server_action/models/fsm_order.py
```diff
@@ -0,0 +1,31 @@
...
+            context = {
+                "active_model": self._name,
+                "active_ids": [res.id],
+            }
+            action.with_context(**context).run()
+        return res
+
+    def write(self, vals):
+        res = super().write(vals)
+        action = self.env["fsm.stage"].browse(vals.get("stage_id")).action_id
```
yankinmax:
```suggestion
        for record in self:
            action = self.env["fsm.stage"].browse(record.stage_id).action_id
```

yankinmax (commented):
Loop over records not vals

max3903 (approved):

max3903:
@BojanOD It would be nice to have unit tests. Thanks.

BojanOD:
> @BojanOD It would be nice to have unit tests. Thanks.

@max3903 I have added the unit tests.

fieldservice_stage_server_action/models/fsm_equipment.py
```diff
@@ -0,0 +1,33 @@
...
+
+class FSMEquipment(models.Model):
+    _inherit = "fsm.equipment"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        for record, vals in zip(records, vals_list):
+            action = record.stage_id.action_id
+            if vals.get("stage_id") and action:
```
ivantodorovich:
```suggestion
        for record in records:
            action = record.stage_id.action_id
            if action:
```

Checking for `vals['stage_id']` seems unnecessary here.
As you're creating records, you don't need to check vals for updates.. simply see if you have an action on the stage_id.

fieldservice_stage_server_action/models/fsm_location.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo import api, models
+
+
+class FSMLocation(models.Model):
+    _inherit = "fsm.location"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        for record, vals in zip(records, vals_list):
```
ivantodorovich: Same comment than before

fieldservice_stage_server_action/models/fsm_order.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo import api, models
+
+
+class FSMOrder(models.Model):
+    _inherit = "fsm.order"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        for record, vals in zip(records, vals_list):
```
ivantodorovich: Same comment than before

fieldservice_stage_server_action/models/fsm_order.py
ivantodorovich: Actually I think currently it's better

Currently, if `stage_id` is not in `vals`, `vals.get("stage_id")` would be `None`.
And thus `action = self.env["fsm.stage"].browse(None).action_id` would be an empty recordset.

This is expected, as we don't want to trigger any action if `stage_id` wasn't updated.

----

If we don't read `stage_id` from vals, you'd be triggering the action for any write. That is, even if some other random field was updated, and not the stage. This is not desired.


----

I do think, though, this piece of code is a bit difficult to read due to the fact that you rely on `vals.get("stage_id")` to be `None` if it's missing, and that `.browse(None)` returns an empty recordset.

I would prefer somethingl like this instead, to make it more explicit:

```py
    def write(self, vals):
        res = super().write(vals)
        action = self.env["fsm.stage"].browse(vals["stage_id"]).action_id if "stage_id" in vals else None
```

or

```py
    def write(self, vals):
        res = super().write(vals)
        if "stage_id" in vals:
            action = self.env["fsm.stage"].browse(vals.["stage_id"]).action_id
            # ...
```

fieldservice_stage_server_action/models/fsm_person.py
```diff
@@ -0,0 +1,33 @@
...
+from odoo import api, models
+
+
+class FSMPerson(models.Model):
+    _inherit = "fsm.person"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        for record, vals in zip(records, vals_list):
```
ivantodorovich: Same comment than before

fieldservice_stage_server_action/tests/test_fsm_stage.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests.common import TransactionCase
+
+
+class FSMRecurringCase(TransactionCase):
+    def setUp(self):
```
ivantodorovich: Use `setUpClass(cls)` instead, as it would be more performant.

`setUp` runs before every test case
`setUpClass` runs once for all of them

ivantodorovich (approved):

Issue #1025, opened 2022-12-29
[14.0][IMP] fieldservice_maintenance: create maintenance request but no fsm_order instead of error


PicchiSeba:
**Is your feature request related to a problem?**
I am currently having this situation:
I'm using `fieldservice` and `maintenance` modules, with `fieldservice_maintenance` to make them work together.
There is some equipment assigned to various locations and other equipment that are not assigned to any location in particular.
When it's time to create the second type of equipment's maintenance_request the operation is stopped because fsm_orders cannot be created in the absence of a location.
I don't care about the fsm_order, I just need a maintenance request so that those responsible can fix it.

**Describe the solution you'd like**
A possible solution for all the equipment without a location could be to let create the maintenance request but not the order and then log the event with the corresponding `fsm_equipment.id` to inform the administrator.

Something similar to this:
```py
if fsm_equipment.current_location_id.id:
   fsm_order_id = self.env["fsm.order"].create(
      {
         "type": fsm_order_type.id,
         "equipment_id": fsm_equipment.id,
         "location_id": fsm_equipment.current_location_id.id,
         "request_id": request.id,
         "description": request.description,
         "request_early": request.schedule_date,
         "scheduled_date_start": request.schedule_date,
         "priority": request.priority,
      }
   )
   request.fsm_order_id = fsm_order_id
else:
   _logger.info(
      _("Missing location on fsm_equipment %s. fsm_order not created" % fsm_equipment.id)
   )
```
by doing so the exception previously blocking the chain of execution does not exist anymore and

**Describe alternatives you've considered**
I have considered the creation of a location whose sole purpose is to represent a non-existing place for the effected equipment but I think that it would just make more confusion.

PR #1026, opened 2022-12-29
[14.0][IMP] fieldservice_maintenance: maintenance request is created …

PicchiSeba:
…even if location_id is not set. But fsm_order is not generated.

Hi,
this PR references issue #1025.

This is the solution that i propose to fix the situation that is affecting me.
I believe it could be useful to others as well.

EDIT: spelling

### DIFF
--- a/fieldservice_maintenance/models/maintenance_request.py
+++ b/fieldservice_maintenance/models/maintenance_request.py
@@ -2,7 +2,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
-from odoo.exceptions import UserError
 
 
 class MaintenanceRequest(models.Model):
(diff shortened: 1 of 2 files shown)


PicchiSeba:
I just rebased the relative commit on the most recent 14.0 version

hparfr:
What do you think of having a flag `create fsm.order` on maintenance requests ? 
Because, you may not want fsm.order at all even if there is a location attached.

PicchiSeba:
I will look into it and let you know

fieldservice_maintenance/models/maintenance_request.py
```diff
@@ -24,23 +28,26 @@ def create(self, vals):
...
-                }
-            )
-
-            request.fsm_order_id = fsm_order_id
+            if request.create_fsm_order:
+                if not fsm_equipment.current_location_id.id:
+                    _logger.info(
+                        _(
+                            "Missing location on fsm_equipment %s. fsm_order was not created"
+                            % fsm_equipment.id
```
max3903: @PicchiSeba What do you think of adding this message to the chatter of the maintenance request as well? The user does not see the logs.

PicchiSeba: makes sense. thank you

PicchiSeba: Hi, sorry for being late. I applied the suggestion

max3903 (approved):

PicchiSeba:
I just updated the relative test file

PicchiSeba:
I'm sorry for the many force pushes, but I don't understand what's the matter with codecov.

The pipeline error says that i decreased code coverage by 0.01% but checking on codecov's website I see 100% coverage.
https://app.codecov.io/gh/OCA/field-service/commit/5f90535da3a67d4ffa82a00ec12bb23073bb35d1
Am I reading that correctly or am I missing something?

PR #1029, opened 2023-01-04
[14.0][IMP]fieldservice_activity: duplicate fsm.order activities

HekkiMelody:
1. When duplicating a FS Order, also duplicate related activities, but reset their state to "todo"
2. When creating a new fsm.order, only copy the activities from the template if none have been already added, do not blindly overwrite them.
3. Trigger the onchange function on an fsm.order's template_id to overwrite existing activities only when adding a template, do not empty out the activities when removing it.

### DIFF
--- a/fieldservice_activity/models/fsm_order.py
+++ b/fieldservice_activity/models/fsm_order.py
@@ -9,38 +9,40 @@ class FSMOrder(models.Model):
     _inherit = "fsm.order"
 
     order_activity_ids = fields.One2many(
-        "fsm.activity", "fsm_order_id", "Order Activities"
+        comodel_name="fsm.activity",
+        inverse_name="fsm_order_id",
+        string="Activities",
+        copy=True,
     )
 
     @api.onchange("template_id")
     def _onchange_template_id(self):
         res = super()._onchange_template_id()
-        for rec in self:
+        for rec in self.filtered("template_id"):
             # Clear existing activities
-            rec.order_activity_ids = [(5, 0, 0)]
-            if rec.template_id:
-                activity_list = []
-                for temp_activity in rec.template_id.temp_activity_ids:
-                    activity_list.append(
-                        (
-                            0,
-                            0,
-                            {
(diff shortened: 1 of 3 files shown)


PicchiSeba:
Rebased on most recent 14.0 version

max3903 (approved):

max3903:
@PicchiSeba Can you add unit tests?

HekkiMelody:
can this be merged? Tests were added

fieldservice_activity/models/fsm_order.py
```diff
@@ -9,38 +9,40 @@ class FSMOrder(models.Model):
     _inherit = "fsm.order"
 
     order_activity_ids = fields.One2many(
-        "fsm.activity", "fsm_order_id", "Order Activities"
+        comodel_name="fsm.activity",
+        inverse_name="fsm_order_id",
+        string="Activites",
```
max3903:
```suggestion
        string="Activities",
```

PicchiSeba:
I applied the suggestion and rebased on the most recent commit (c177ae7a2f9a44b0160dccdfd65ce45232d2da05)

PicchiSeba:
The error seems to be unrelated this this issue in particular
```
2023-07-03 14:22:13,169 251 ERROR odoo odoo.addons.fieldservice_geoengine.tests.test_fsm_location: FAIL: TestFsmLocation.test_fsm_location_creation
  File "/__w/field-service/field-service/fieldservice_geoengine/tests/test_fsm_location.py", line 63, in test_fsm_location_creation
    test_location_1.partner_latitude, 50.629850, delta=self.delta
AssertionError: 50.62998 != 50.62985 within 1e-05 delta
```

That test in particular was changed here #1097 because of a problem that firstly appeared (after a while) here #1039 

In #1039 @hparfr suggested
>Yes but, is it the responsibility of this module to test the trueness of this adress lookup ?

Would it make sense to actually skip that test altogether?

PR #1031, opened 2023-01-04
[14.0][FIX]fieldservice_recurring: fix security for group_fsm_user_own

HekkiMelody:
Fixes to let users without access to recurring orders (most notably, `group_fsm_user_own` still access and use the fieldservice app without permission errors.


max3903 (approved):

hparfr (approved):
/ocabot merge minor

PR #1032, opened 2023-01-06
[14.0][IMP] fieldservice_sale : new selection option, Create multiple FSM orders per sale order

chandni299:
Use Case:
On `product.template`
name: Product1
field_service_tracking: sale_multiple
fsm_order_template_ids: template1, template2

On `sale.order`
order_line1 : Product1

Result: 
Then two FSM orders will be created for one sale order as we have assigned two templates on Product1


hparfr:
What's the use case ?

max3903:
@hparfr Use case I can think of: In the construction industry, you sell one thing but you need to come on site multiple times to complete the delivery/installation. Electricity for example: you need to come first when the walls are open to wire the house, you then come back to install the electric panel and connect the cable, finally when walls are painted, you install the lights and interruptors.

max3903:
@chandni299 Can you increase test coverage a bit?

hparfr:
Having multiple fsm orders tied directly to one sale.order will add a lot of complexity in the module.

There is other better designs:
- add sale order lines to support all the fsm.orders -> easy to invoice the right amount  / partial invoincing
- use fsm_sale_recurring
- implement a kind of backorders like in stock module

benbrich:
The use case @chandni299 and I are trying to cover is from the rental domain. The respective company is renting out physical products as a service at a field location. The products need to be assembled and disassemled in the field and all is sold as one service with only one sale order line per physical product. @hparfr using multiple lines is not an option due to customer restrictions and I do not see how to cover this case using fsm_sale_recurring either. Can you tell a little bit more about your suggestion with the backorder please?

PR #1036, opened 2023-01-13
[14.0][FIX] fieldservice_geoengine: fsm_user_own can see maps

PicchiSeba:
I also cleaned up redundant settings


HekkiMelody:
Test failures don't seem related to the changes. I see the same test failures pop up everywhere, even after translations...

Just two from the history:
https://github.com/OCA/field-service/actions/runs/3884911196/jobs/6628097775
https://github.com/OCA/field-service/actions/runs/3752618645/jobs/6374980914

@OCA/field-service-maintainers

max3903 (approved):

PR #1039, opened 2023-01-20
[14.0][IMP] fielservice_geoengine: refactor of fieldservice_geoengine module

PicchiSeba:
WARNING: this PR was made some time ago. I fixed some things and had to rebase on the most recent commit.

I would like to propose some improvements to the fieldservice_geoengine module that we have been using and that we think could be helpful to the OCA field-service community.

fsm.order has been changed to be related to fsm.location with a many2one. This removes the need for a local shape, using an existing location instead. By doing so some methods are not needed anymore because they are in fsm.location already.

fsm.location has also been subject to changes:
 - some missing fields have been added.
 - shape attribute has been altered to a computed field in order to avoid unnecessary recalculations when it's requested. It's recomputed only when partner's latitude and/or longitude are changed.

The map view previously added by this module in the fsm_order form view has also been made available in fsm_location form view. This change was made so that a user does not have to generate an fsm_order in order to see the map with the fsm_location's position.

In addition to the previous changes the tests have been expanded for a broader control.

### DIFF
--- a/fieldservice_geoengine/tests/test_fsm_location.py
+++ b/fieldservice_geoengine/tests/test_fsm_location.py
@@ -1,52 +1,154 @@
 # Copyright (C) 2012 - TODAY, Open Source Integrators
+# Copyright (C) 2023 - TODAY Pytech SRL
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import fields
-from odoo.tests.common import TransactionCase
+from odoo.tests.common import SavepointCase
 
 
-class TestFsmLocation(TransactionCase):
-    def setUp(self):
-        super(TestFsmLocation, self).setUp()
-        self.fsm_location = self.env["fsm.location"]
-        self.location_partner_1 = self.env.ref("fieldservice.location_partner_1")
-        self.location_partner_2 = self.env.ref("fieldservice.location_partner_2")
-        self.test_loc_partner = self.env.ref("fieldservice.test_loc_partner")
-        self.location_partner_3 = self.env.ref("fieldservice.location_partner_3")
+class TestFsmLocation(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.FSMLocation = cls.env["fsm.location"]
+        cls.location_partner_1 = cls.env.ref("fieldservice.location_partner_1")
+        cls.location_partner_2 = cls.env.ref("fieldservice.location_partner_2")
+        cls.location_partner_3 = cls.env.ref("fieldservice.location_partner_3")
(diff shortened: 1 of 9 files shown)


max3903 (approved):

fieldservice_geoengine/tests/test_fsm_location.py
```diff
@@ -1,52 +1,154 @@
...
+                "phone": "123",
+                "email": "test@example.com",
+                "partner_id": test_partner.id,
+                "owner_id": self.location_partner_1.id,
+            }
+        )
+        self.assertTrue(test_location_1.partner_latitude)
+        self.assertTrue(test_location_1.partner_longitude)
+        self.assertAlmostEqual(
+            test_location_1.partner_latitude, 50.629981, delta=self.delta
```
hparfr, line 63: now, tests are failling on 14.0 because the value is too precise
 
```
 2023-05-17 09:50:34,282 254 ERROR odoo odoo.addons.fieldservice_geoengine.tests.test_fsm_location: FAIL: TestFsmLocation.test_fsm_location_creation
  File "/__w/field-service/field-service/fieldservice_geoengine/tests/test_fsm_location.py", line 63, in test_fsm_location_creation
    test_location_1.partner_latitude, 50.629981, delta=self.delta
AssertionError: 50.629850000000005 != 50.629981 within 1e-05 delta
``` 
like https://github.com/OCA/field-service/actions/runs/5001740733/jobs/8960810974#step:8:167


both values looks legit 

@PicchiSeba  do you know why the results are not consistent  ?
can we reduce the delta to like 0,001 or may write a different test

PicchiSeba: Is this an isolated test failure or has it happened before?
We can loosen the precision from 0.000_01 to 0.000_1 to temporarily fix the pipelines while we investigate the reason of these failures.

I think loosening it more than that would be bad to check the correct module functionality.
That digit we are skipping (from 5 to 4) means checking for precision of ~10 metres instead of ~1 meter.

hparfr: Each commit on 14.0 has this error (like translation updates).

>I think loosening it more than that would be bad to check the correct module functionality.

Yes but, is it the responsibility of this module to test the trueness of this adress lookup ?

PicchiSeba: That's a good point actually.

This module should just integrate base_geoengine into fieldservice. It is not this module's responsibility to actually check the correctness of all the geolocation calculations

hparfr: fixed  in https://github.com/OCA/field-service/pull/1097

PR #993, opened 2022-11-03
[16.0][MIG] fieldservice_stock

ghost:

PR #1041, opened 2023-01-30
[16.0][MIG] fieldservice_sale

epanisset:


fieldservice_sale/models/sale_order.py
```diff
@@ -41,25 +39,25 @@ def onchange_partner_id(self):
...
-        self.fsm_location_id = location_ids and location_ids[0] or False
+        location = self.partner_id.fsm_location
+        if not location:
+            domain = [
+                "|",
+                "|",
+                ("partner_id", "=", self.partner_id.id),
+                ("partner_id", "=", self.partner_shipping_id.id),
+                ("partner_id", "=", self.partner_id.commercial_partner_id.id),
+            ]
```
yankinmax: can you pls use OR expression from osv?
```
expression.OR(
    [
        [("partner_id", "=", self.partner_id.id)],
        [("partner_id", "=", self.partner_shipping_id.id)],
        [("partner_id", "=", self.partner_id.commercial_partner_id.id)],
    ]
)
```

yankinmax:
@SilvioC2C this can cause Travis failure not related to this module
`ERROR odoo odoo.schema: Table 'fsm_location': unable to set NOT NULL on column 'inventory_location_id'`
I suggest to open one more PR to set required attribute directly in view for this field.
If not - LG

SilvioC2C:
@yankinmax that shouldn't cause Travis to fail; moreover, that field is defined in `fieldservice_stock`, not `fieldservice_sale`.

yankinmax:
@SilvioC2C I completely agree with you, but modules are installed altogether in the repo, so I had such situations.

SilvioC2C:
@yankinmax that's ok, but you should open an issue for that or mention this in the proper PR, this is the wrong place :smile:

yankinmax:
@SilvioC2C I've opened a request on the migration PR for `fieldservice_stock`:
https://github.com/OCA/field-service/pull/993#issuecomment-1410106116

fieldservice_sale/models/sale_order.py
SilvioC2C: Why? We're not composing domain dinamically, so we know beforehand which conditions are to be chained with an `OR`, so we can create the domain directly using the `"|"` operator. It's easier to read, and doesn't force Odoo to compute a new domain.

SilvioC2C (approved):

brian10048:
Would you be able to add #1050 in this migration?

Camille0907 (approved):
LGTM

dalonsod:
@epanisset would you resume this PR? In that case, please add #1072 if needed. Thanks!

stefan-tecnativa (approved):
Functionally tested 👍️. @epanisset what is the state of this PR?

pedrobaeza:
Superseded by #1101, but respecting commit attribution.

PR #1057, opened 2023-03-17
[16.0][IMP] fieldservice: Add name for the different pages in fsm_location.xml to a…

FranzPoize:
…llow xpath selection


max3903 (approved):

FranzPoize:
I evidently made a mistake and want to put this on 14.0 should I make another PR for the 14.0 and also keep this one or can we do this another way

PR #1059, opened 2023-03-31
[16.0][IMP] fieldservice: improvements for scheduled dates

marcelsavegnago:
Fix scheduled_date_end update when duration is zero and use record value for unchanged duration

This commit addresses an issue in the calculation of scheduled dates, where the end date was not being updated correctly when the duration was zero. Additionally, the _calc_scheduled_dates function now uses the current value in the record when the duration is not changed. The end date is now updated properly, regardless of the duration value.

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -272,7 +272,7 @@ class FSMOrder(models.Model):
         """Calculate scheduled dates and duration"""
 
         if (
-            vals.get("scheduled_duration")
+            vals.get("scheduled_duration") is not None
             or vals.get("scheduled_date_start")
             or vals.get("scheduled_date_end")
         ):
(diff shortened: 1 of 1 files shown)


fieldservice/models/fsm_order.py
```diff
@@ -328,6 +328,10 @@ def action_cancel(self):
             {"stage_id": self.env.ref("fieldservice.fsm_stage_cancelled").id}
         )
 
+    @api.onchange("scheduled_date_start")
```
max3903: @marcelsavegnago Why not adding this field in the decorator of `onchange_scheduled_duration()` directly?

marcelsavegnago: Done

max3903 (approved):

rvalyi (approved):

PR #1068, opened 2023-04-17
[15.0][FIX] fieldservice_stock: default inventory location not set

dalonsod:
Before this fix, when creating a new FSM location inventory location default value was reset by parent location onchange method, when parent location was initially unset.
After this commit, only when parent location is set inventory location could be updated, preveting the described bug.

### DIFF
--- a/fieldservice_stock/models/fsm_location.py
+++ b/fieldservice_stock/models/fsm_location.py
@@ -18,5 +18,6 @@ class FSMLocation(models.Model):
     @api.onchange("fsm_parent_id")
     def _onchange_fsm_parent_id(self):
         res = super(FSMLocation, self)._onchange_fsm_parent_id()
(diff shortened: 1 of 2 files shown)


max3903 (approved):

dalonsod:
@vimalpatelserpentcs could you review? Thanks!

brian10048 (approved):

dalonsod:
@max3903 I've noticed that pre-commit (and codecov) failed during merge, but in PR actions is fine

max3903:
@dalonsod It could be due to a new version of node. Can you push a dummy commit to trigger Github Actions one more time?

dalonsod:
Well, pre-commit is still green, but test coverage is now in red, I'll take a look

dalonsod:
@max3903 could you try again?

PR #1069, opened 2023-04-19
[15.0][FIX] fieldservice: replace type Text to Html in todo field

ChristianSantamaria:
Due to the version change from 14.0 to 15.0, the t-raw was replaced by t-out, this field was not displayed correctly in the service orders report, an attempt was made to change the t-out to others such as t-esc or t- field. but it is not displayed correctly, then choose to change the type to html, since in all the views that are called this field the widget is added.

Before in runboat with Text field:
imagen

After with Html field:
imagen

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -147,7 +147,7 @@ class FSMOrder(models.Model):
     scheduled_duration = fields.Float(help="Scheduled duration of the work in" " hours")
     scheduled_date_end = fields.Datetime(string="Scheduled End")
     sequence = fields.Integer(default=10)
-    todo = fields.Text(string="Instructions")
+    todo = fields.Html(string="Instructions")
 
     # Execution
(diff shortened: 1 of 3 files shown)


max3903 (approved):

max3903:
@ChristianSantamaria Not sure how this change impacts the data. Do we need a migration script to convert from text to html?

ChristianSantamaria:
@max3903 
I don't think it's necessary, I've been doing tests previously without this change, that is to say with the field of type Text and save. As in all the views in which this field appears, it contains the html widget because the same is stored in the database, here are some examples:
imagen

And when updating the addon, the values in the database do not change as you can see:
imagen

I also have to solve the issue of the tests.

ChristianSantamaria:
@max3903 Done 👍

dalonsod (approved):
LGTM

ChristianSantamaria:
@max3903 all ready, can you merge it?

PR #1088, opened 2023-05-10
[14.0][FIX] fieldservice_account_analytic: Write correct value for customer_id

michelerusti:
The value for customer_id was written with the wrong value as the `order.location_id.customer_id` points to a different value than the sale order that generates the fsm_order

Also, it was not written at creation

### DIFF
--- a/fieldservice_account_analytic/models/fsm_order.py
+++ b/fieldservice_account_analytic/models/fsm_order.py
@@ -56,8 +56,7 @@ class FSMOrder(models.Model):
             self.location_id = self.customer_id.service_location_id
 
     def write(self, vals):
-        res = super(FSMOrder, self).write(vals)
         for order in self:
-            if "customer_id" not in vals and order.customer_id is False:
-                order.customer_id = order.location_id.customer_id.id
-        return res
+            if "customer_id" not in vals and not order.customer_id:
+                vals.update({"customer_id": order.location_id.customer_id.id})
+        return super(FSMOrder, self).write(vals)
--- a/fieldservice_sale/models/fsm_order.py
+++ b/fieldservice_sale/models/fsm_order.py
@@ -20,3 +20,15 @@ class FSMOrder(models.Model):
             "context": {"create": False},
             "name": _("Sales Orders"),
         }
+
+    def write(self, vals):
+        for order in self:
+            if "customer_id" not in vals and not order.customer_id:
+                vals.update({"customer_id": order.sale_id.partner_id.id})
+        return super(FSMOrder, self).write(vals)
+
+    def create(self, vals):
+        sale_id = self.env["sale.order"].browse(vals.get("sale_id"))
+        if sale_id:
+            vals["customer_id"] = sale_id.partner_id.id
+        return super(FSMOrder, self).create(vals)
--- a/fieldservice_sale/models/fsm_order.py
+++ b/fieldservice_sale/models/fsm_order.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2019 Brian McMaster
 # Copyright (C) 2019 Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 
 
 class FSMOrder(models.Model):
@@ -27,6 +27,7 @@ class FSMOrder(models.Model):
                 vals.update({"customer_id": order.sale_id.partner_id.id})
         return super(FSMOrder, self).write(vals)
 
+    @api.model
     def create(self, vals):
         sale_id = self.env["sale.order"].browse(vals.get("sale_id"))
         if sale_id:
--- a/fieldservice_stock_account/tests/test_fsm_stock_account.py
+++ b/fieldservice_stock_account/tests/test_fsm_stock_account.py
@@ -3,6 +3,7 @@
 
 import datetime
 
+from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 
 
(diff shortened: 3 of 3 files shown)


max3903:
@michelerusti If you could set those fields in vals before calling super(), they would be set as part of the same transactions.

michelerusti:
> @michelerusti If you could set those fields in vals before calling super(), they would be set as part of the same transactions.

Yes, thanks for the advise
I've also removed the error assertion in `test_fsm_stock_account.py` because with this fix the user is set correctly

Let me know if there's anything else needed to do!

fieldservice_account_analytic/models/fsm_order.py
```diff
@@ -58,6 +58,6 @@ def _onchange_customer_id_location(self):
     def write(self, vals):
         res = super(FSMOrder, self).write(vals)
         for order in self:
-            if "customer_id" not in vals and order.customer_id is False:
-                order.customer_id = order.location_id.customer_id.id
+            if "customer_id" not in vals and not order.customer_id:
+                order.customer_id = order.location_id.customer_id
```
max3903: @michelerusti I meant something like this:

```
    def write(self, vals):
        if "customer_id" not in vals and not self.customer_id:
            vals.update({"customer_id": self.location_id.customer_id.id})
        res = super().write(vals)
```

michelerusti: @max3903 I've edited,  me know if it's alright!

fieldservice_sale/models/fsm_order.py
```diff
@@ -20,3 +20,16 @@ def action_view_sales(self):
             "context": {"create": False},
             "name": _("Sales Orders"),
         }
+
+    def write(self, vals):
+        res = super(FSMOrder, self).write(vals)
+        for order in self:
+            if "customer_id" not in vals and not order.customer_id:
+                order.customer_id = order.sale_id.partner_id
+        return res
```
max3903: @michelerusti Same idea for this method

fieldservice_sale/models/fsm_order.py
```diff
@@ -20,3 +20,16 @@ def action_view_sales(self):
...
+
+    def create(self, vals):
+        res = super(FSMOrder, self).create(vals)
+        for order in res:
+            order.customer_id = order.sale_id.partner_id
+        return res
```
max3903: @michelerusti Same idea but it's a bit different: you need to search the sale order with vals.get("sale_id") to get partner_id first. 1 select and 1 insert is still better than 1 insert and 1 update.

max3903 (approved):

fieldservice_sale/models/fsm_order.py
```diff
@@ -20,3 +20,15 @@ def action_view_sales(self):
             "context": {"create": False},
             "name": _("Sales Orders"),
         }
+
+    def write(self, vals):
+        for order in self:
+            if "customer_id" not in vals and not order.customer_id:
```
hparfr, line 26: `[fsm.order]customer_id` is not in the dependencies.

fieldservice_sale/models/fsm_order.py
```diff
@@ -20,3 +20,15 @@ def action_view_sales(self):
...
+                vals.update({"customer_id": order.sale_id.partner_id.id})
+        return super(FSMOrder, self).write(vals)
+
+    def create(self, vals):
+        sale_id = self.env["sale.order"].browse(vals.get("sale_id"))
```
hparfr, line 32: this should be inherit from here : https://github.com/OCA/field-service/blob/14.0/fieldservice_sale/models/sale_order.py#L67

hparfr (changes requested):

fieldservice_sale/models/fsm_order.py
michelerusti: I see... 
Since the field is defined in fieldservice_accont_analytic, how would you consider the idea to add that module as a dependency?
Otherwise all this PR would loose meaning

Let me know if there is any different approach you would've preferred

hparfr: There is different approach here. The most two obvious are: 

-  you use fieldservice_sale then you may want to invoice based on the sale order (and use analytics from sale order lines)
-  you use create orders directly so you don't have a sale, thanks to fieldservice_account_analytic,  you invoice the fsm orders directly.


Do you use another approach ? 

From a technical perspective, you can put your code in a new module like `fieldservice_sale_analytic`.

michelerusti: We create the sale order and let the fsm order be generated by `fieldservice_sale` . 
The problem is the `customer_id` that results empty when it should be the same customer of the sale order 

This means that `customer_id` should always be available to `fieldervice_sale` since otherwise it breaks the module and the options are re-difining `customer_id` or add the dependency of `fieldservice_sale_analytic` . 

I see this as a bug since otherwise the `customer_id` is not automatically set. 
Let me know your thoughts about it and how you think is best to proceed, or even if it's to be considered a bug at all!

fieldservice_sale/models/fsm_order.py
michelerusti: Inherited... how?
Looking in the code, the method `_prepare_fsm_values()` is not called by the `create()` but from other methods [like this one](https://github.com/OCA/field-service/blob/880c8076f04716d55bbaca9ecba10217672fb64b/fieldservice_sale/models/sale_order.py#L109)

fieldservice_sale/models/fsm_order.py
michelerusti: > From a technical perspective, you can put your code in a new module like `fieldservice_sale_analytic`.

do you think that for this approach it needs to be created a new module?

hparfr: If you use sale to generate the fsm order, you do not need customer_id. The order will be invoiced directly by the sale.

Why using fieldservice_analytic ?

michelerusti: > If you use sale to generate the fsm order, you do not need customer_id.

@hparfr 
And is this definitive or is it open to discussion? It may not be essential but I think that having the correct customer reference would be nice

PR #1091, opened 2023-05-17
[15.0][FIX] fieldservice_sale: fsm_location_id is mandatory only if order line display_type not in ('line_section', 'line_note')

yankinmax:
We should be able to confirm Sale Order if `fsm_location_id` is not set and the order line is Note or Section.

### DIFF
--- a/fieldservice_sale/tests/test_fsm_sale_order.py
+++ b/fieldservice_sale/tests/test_fsm_sale_order.py
@@ -487,3 +487,45 @@ class TestFSMSaleOrder(TestFSMSale):
         self.assertEqual(
             len(invoices.ids), 1, "FSM Sale: Sale Order 4 should create 1 invoice"
         )
+
+    def test_sale_order_5(self):
+        """Test ValidationError isn't raised if order line
+        display_type in ("line_section", "line_note")
+        """
+        # remove normal order line with display_type=False
+        self.sol_service_per_order.unlink()
+        # add note as order line to sale order
+        self.sol_note = self.env["sale.order.line"].create(
+            {
+                "name": "This is a note",
+                "display_type": "line_note",
+                "product_id": False,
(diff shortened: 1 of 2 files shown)


max3903 (approved):

gurneyalex (approved):

fieldservice_sale/tests/test_fsm_sale_order.py
```diff
@@ -487,3 +487,24 @@ def test_sale_order_4(self):
...
+                "product_uom_qty": 0,
+                "product_uom": False,
+                "price_unit": 0,
+                "order_id": self.sale_order.id,
+                "tax_id": False,
+            }
+        )
+        # Confirm the sale order
+        with self.assertRaises(ValidationError), self.cr.savepoint():
+            self.sale_order.action_confirm()
```
SilvioC2C: This test is quite confusing: the docstring says it will test that the `ValidationError` should be raised only when `display_type not in ("line_section", "line_note")`, but then the test proceeds to add a `line_note` type sale line, and a `ValidationError` is raised anyway.
I'd rewrite this test by trying to confirm 3 orders without `fsm_location_id`:
1. the first one with only a normal sale line => `ValidationError` should be raised
2. the second one with only a `line_section` sale line => `ValidationError` should not be raised
3. the third one with only a `line_note` sale line => `ValidationError` should not be raised

yankinmax: `the first one with only a normal sale line => ValidationError should be raised`
this is already tested in `test_sale_order_0`
so I would go with 2 and 3 option, thanks

yankinmax: @SilvioC2C I updated tests :smile:

SilvioC2C (approved):

PR #1112, opened 2023-08-13
[IMP][16.0] fieldservice: improved views concerning categories & tags

NICO-SOLUTIONS:


max3903 (approved):

fieldservice/views/fsm_location.xml
```diff
@@ -9,6 +9,7 @@
                 <field name="complete_name" />
                 <field
                     name="category_id"
+                    string="Partner-Tags"
```
hparfr: question: why this special handling for this field ?

there is other fields (street, mobile, etc.) from the partner too.  Specify "partner" only on this field can create confusion for other fields

hparfr (changes requested):
Thanks for these improvments.

I have remarks about Partner-Tags.
I'm ok with the rest of the proposed changes.

fieldservice/views/fsm_location.xml
NICO-SOLUTIONS: i added this string in order to separate it more clearly from the category_id that is used for Worker Categories (Skills).
If there is no clear need, i will remove it. What do you think? @hparfr

hparfr: Yes please

NICO-SOLUTIONS: done

hparfr (approved):
Thanks @NICO-SOLUTIONS

PR #1118, opened 2023-08-31
[14.0] [IMP] fieldservice_isp_account: decouple move vals preparation from actual records creation

are-agilebg:
This PR proposes to split the methods that are responsible for the creation of account moves (create_bills and account_create_invoice) both into two methods, one that only prepares the data (vals) and one that effectively create the record.

This way there is a clean way to hook into move creation logic for modules that depends on fieldservice_isp_account.


tafaRU (approved):

tafaRU:
@max3903 and what about this? Could you please review it?

max3903 (approved):

tafaRU:
thanks @max3903!

PR #1122, opened 2023-10-25
[FIX] fieldservice_sale_*

brian10048:
Do not copy FSM Order or FSM Recurring Orders on sale order line

### Steps to Reproduce:
- Create a sale order using a product which has field service tracking set to order (or recurring order).
- Confirm sale order
- Duplicate sale order
- Confirm the duplicated sale order

### Expected Result:
The duplicated sale order should create new field service order (or recurring order)

### Actual Result:
No field service order (or recurring order) is created because the new order lines in duplicate sale order already have a reference to the previous order lines' fsm_order_id


max3903 (approved):

hparfr (approved):
/ocabot merge minor

Issue #963, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [x] base_territory by @max3903 in #966
- [x] fieldservice by @max3903 in #967
- [x] fieldservice_account - By @gbrito-erpopen - #985
- [x] fieldservice_account_analytic - By @juliette-blanc - #1084 & @michelerusti #1127 
- [x] fieldservice_account_payment - By @rrebollo - #1489
- [x] fieldservice_activity - By @gbrito-erpopen - #982
- [ ] fieldservice_agreement #1130 
- [x] fieldservice_calendar - By @juliette-blanc - #1074
- [ ] fieldservice_change_management - By @aiendry-aktivsoftware - #1099
- [x] fieldservice_crm - By @gbrito-erpopen - #983
- [ ] fieldservice_delivery
- [ ] fieldservice_distribution
- [x] fieldservice_equipment_stock - By @szekawong - #1129
- [x] fieldservice_fleet - By @rrebollo - #1491
- [x] fieldservice_geoengine - By @peluko00 - #1169
- [ ] fieldservice_google_map
- [ ] fieldservice_google_marker_icon_picker
- [x] fieldservice_isp_account #1126 
- [x] fieldservice_isp_flow - By @peluko00 - #1168
- [ ] fieldservice_location_builder
- [ ] fieldservice_maintenance - By @jasperjumelet - #1056
- [ ] fieldservice_partner_multi_relation
- [x] fieldservice_portal - By @anusriNPS - #1224
- [x] fieldservice_project by @epanisset #1045 
- [ ] fieldservice_purchase
- [x] fieldservice_recurring - By @kaynnan - #1105
- [ ] fieldservice_repair
- [x] fieldservice_route - By @gbrito-erpopen - #992
- [ ] fieldservice_route_account
- [ ] fieldservice_route_stock
- [ ] fieldservice_route_vehicle
- [x] fieldservice_sale - By @stefan-tecnativa - #1101
- [x] fieldservice_sale_recurring - #1149 
- [ ] fieldservice_sale_stock
- [x] fieldservice_size - By @kaynnan - #1106
- [x] fieldservice_skill - By @stefan-tecnativa - #1094
- [ ] fieldservice_stage_server_action
- [x] fieldservice_stage_validation - By @juliette-blanc - #1082
[...]

PR #1126, opened 2023-11-01
[16.0][MIG] fieldservice_isp_account

michelerusti:
migration for #963 

Depends on:

- [x] https://github.com/OCA/field-service/pull/1127


michelerusti:
There are a bunch of errors regarding wrong INSERT and `not-null constraint` like this one:
[https://github.com/OCA/field-service/actions/runs/6851350407/job/18627432083?pr=1126#step:8:628](https://github.com/OCA/field-service/actions/runs/6851350407/job/18627432083?pr=1126#step:8:628)
`psycopg2.errors.NotNullViolation: null value in column "sale_line_warn" violates not-null constraint
DETAIL: ....
`


 
After some analysis, I conclude that those are not actually module related errors but are environment-caused. Correct me if I'm wrong please:
The failing tests are (taking the link above as an example) in `fieldservice_project` that errors out for the missing, required field,  `customer_id` that is defined in [fieldservice_account_analytic in this PR](https://github.com/OCA/field-service/pull/1127/files#diff-0f010bbc8160334c11b2a99d4e000e11ece3e80db8f2d20393bfdacec2fc7c55R15) and not at all related to **isp_account** and that shouldn't fail in any case since the `owner_id` is [passed as a value in the create](https://github.com/OCA/field-service/blob/bba0b25df572ed103dd2d3549d66a39f24b4ad92/fieldservice_project/tests/common.py#L13)
This is already suspect since those errors are not present in that PR's tests

After a bit of trial and error, I managed to have the expected result, meaning all the errors are related to the module of this PR

How?
Not installing **fieldservice_account_analytic** in the database initialization stage but in the "Run test" stage
By doing so, the tests pass correctly and I've fixed all the one that required to

Why?
idk, somehow if installed before, the database structure changes accordingly (hence the db errors) but the model and the python code is not loaded correctly. I haven't found out why

As per this PR, all the failing tests right now are to be considered junk. 


@max3903 @brian10048 do you have any consideration about this? Am I right in some way or I'm doing something wrong?
Thanks

stevech091 (approved):
Functional TEST Ok

are-agilebg (approved):

PR #1150, opened 2023-12-19
17.0 [MIG] fieldservice_account

SMaciasOSI:
@max3903 @brian10048

PR #1151, opened 2023-12-19
17.0 [MIG] fieldservice_sale

SMaciasOSI:
@max3903 @brian10048


SMaciasOSI:
This PR requires #1150 to be merged first so the test can pass after a rebase

fieldservice_sale/models/sale_order.py
```diff
@@ -149,15 +149,15 @@ def _field_service_generate(self):
 
         # Process lines set to FSM Sale
         new_fsm_sale_sol = self.order_line.filtered(
-            lambda l: l.product_id.field_service_tracking == "sale"
-            and not l.fsm_order_id
+            lambda L: L.product_id.field_service_tracking == "sale"
+            and not L.fsm_order_id
```
dreispt, line 153: Why the uppercase? Convention is to use lowercase variable names.

SMaciasOSI: > Why the uppercase? Convention is to use lowercase variable names.

With the lowecase I got an error regarding "ambiguous variable name". I got the suggestion to change from l to L from here: https://docs.astral.sh/ruff/rules/ambiguous-variable-name/

PR #1160, opened 2024-02-21
[16.0][FIX] fieldservice_sale: Fix sale order line create method

apatyk:
This PR fixes a bug with the sales order line create method in the fieldservice_sale module to account for cases when the method is triggered without any sales order lines that are created, e.g. on some expense reports. This correction is in line with other `create()` methods and all sales order lines that are created should be returned, not just the last.

Currently, the method throws an undefined variable error for `line` if there are no sales order lines created and only returns the last sales order line if there are any created.

### DIFF
--- a/fieldservice_sale/__manifest__.py
+++ b/fieldservice_sale/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Field Service - Sales",
-    "version": "16.0.1.1.0",
+    "version": "16.0.1.1.1",
     "summary": "Sell field services.",
     "category": "Field Service",
     "author": "Open Source Integrators, Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


houzefa-abba (approved):
Good change, makes perfect sense :D
I'm using other FSM modules but not this specific one so it didn't break anything in my case

I'm surprised such a glaring bug went through, I guess this line creation code is not covered by unit tests? Would be nice to add one around it

Also I think you're not supposed to change the module version in your PR; this is done automatically by a bot when maintainers merge your PR

apatyk:
> Good change, makes perfect sense :D I'm using other FSM modules but not this specific one so it didn't break anything in my case
> 
> I'm surprised such a glaring bug went through, I guess this line creation code is not covered by unit tests? Would be nice to add one around it
> 
> Also I think you're not supposed to change the module version in your PR; this is done automatically by a bot when maintainers merge your PR

@houzefa-abba Awesome, thanks! Just pushed a fix to remove the module version bump.

max3903 (approved):

PR #1164, opened 2024-03-05
[16.0][FIX] fieldservice: Fix location partner type

### LINKED ISSUE #1128: Partner type for FSM locations
Hi, when I create an FSM location on runboat (tried 14.0 & 16.0), the partner created through inheritance has `is_company=false & type=contact`.

I'm surprised it does not use the `type=fsm_location` [added in `res.partner`](https://github.com/OCA/field-service/blob/16.0/fieldservice/models/res_partner.py#L10). I am not sure whether this is a bug or intended.

[Tests](https://github.com/OCA/field-service/blob/16.0/fieldservice/tests/test_fsm_location.py) don't explicitly test partner type when an FSM location is created by hand.

From what I can tell, this `type=fsm_location` is only used when one creates such a partner by hand, which then auto-creates the FSM location (tested in [`test_convert_partner_to_fsm_location`](https://github.com/OCA/field-service/blob/16.0/fieldservice/tests/test_fsm_location.py#L237)).

Should we change so all FSM locations have `type=fsm_location` on their partner?

houzefa-abba:
Service locations `fsm.location` create a partner `res.partner` through inheritance.

Before this change, partners could be explicitely typed "fsm_location" when adding them within children of an existing partner, which in turn would automatically create service locations on top of them.

But creating a service location by hand would not set that type on the partner created through inheritance. After this change, all partners built from service locations are typed the same.

Fixes #1128.

### DIFF
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -70,10 +70,12 @@ class FSMLocation(models.Model):
     )
 
     @api.model_create_multi
-    def create(self, vals):
-        res = super().create(vals)
-        res.write({"fsm_location": True})
-        return res
+    def create(self, vals_list):
+        for vals in vals_list:
+            vals.update({"fsm_location": True, "type": "fsm_location"})
+        return super(FSMLocation, self.with_context(creating_fsm_location=True)).create(
+            vals_list
+        )
 
     @api.depends("partner_id.name", "fsm_parent_id.complete_name", "ref")
     def _compute_complete_name(self):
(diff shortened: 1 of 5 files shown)


max3903 (approved):

max3903:
@houzefa-abba Do you have other changes to push or is this good to be merged?

houzefa-abba:
> @houzefa-abba Do you have other changes to push or is this good to be merged?

Hi this one is good to go, my latest change was only a rebase

PR #1165, opened 2024-03-05
[16.0][IMP] fieldservice: Set location parent to owner

houzefa-abba:
Before this change, service location partners (created through
inheritance) were typed "fsm_location" but did not always have a parent.

When adding them within children of an existing partner, they of course
did have a parent, but not the other way around: they had no parents
when created, through inheritance, upon service location creation.

After this change, service location partners are by default children of
the service location owner partner.

This is only done through defaults but not as related fields, which
would break too much. Therefore, existing partners are not impacted by
this change; only new ones created when creating a service location.

See #1128.

This PR is a follow-up on #1164 where partner types got made more consistent.

### DIFF
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -72,7 +72,10 @@ class FSMLocation(models.Model):
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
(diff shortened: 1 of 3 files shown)


houzefa-abba:
Just rebased since #1164 got merged

This PR is ready

max3903 (approved):

PR #1171, opened 2024-04-04
[16.0][MIG] fieldservice_timeline: Migration to 16.0

houzefa-abba:
**:warning: This migration removes JS code** which was, notably, adding a way to
display available workers & a filter around them.
Noted in readme/ROADMAP.rst.

I tried migrating it but this was leading to too many changes / possibly needing to rewrite some JS code to use owl components.

The timeline view itself works fine, but lacks that available worker list.
In practice, when workers already have other orders assigned to them, this isn't that big of an issue (only workers who have no currently assigned orders will be missing for the worker sidebar).

# Before (v14)
Screenshot at 2024-04-04 18-31-02

# After (v16)
(with https://github.com/OCA/web/pull/2790 applied, otherwise rows would be at the bottom)
Screenshot at 2024-04-04 18-31-48


max3903 (approved):

miquelalzanillas (approved):
Functional tests are OK! Thanks!

PR #1190, opened 2024-05-22
[17.0][MIG] fieldservice_geoengine: Migration to 17.0

peluko00:
Module migrated to version 17.0

cc https://github.com/APSL 155457
@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall please review

- [Issue] https://github.com/OCA/field-service/issues/1132

Depends on:

- [base_geoengine] https://github.com/OCA/geospatial/pull/368


peluko00:
Hi @sersanchus, i have a problem when enter to map view for a problem with source.

image

The problem is in [base_geoengine](https://github.com/APSL/geospatial/blob/17.0-mig-base_geoengine/base_geoengine/static/src/js/views/geoengine/geoengine_renderer/geoengine_renderer.esm.js#L604) that we work together but i don't know why is throwing this error. Any idea to solve that?

Thanks in advance

sersanchus:
Hi @peluko00,

The error is related to this piece of code not finding any vec layer:

`var extent = this.vectorLayersResult
            .find((res) => res.values_.visible === true)
            .getSource()
            .getExtent();`
            
As all openlayers vec layers are visible by default, maybe there aren't any 'geoengine.vector.layer' associated with the geoengine view of your demo. In that case 'getOriginalZoom' has to detect that and use for example an empty extent in taht case:

`ol.extent.createEmpty()`

peluko00:
> Hi @peluko00,
> 
> The error is related to this piece of code not finding any vec layer:
> 
> `var extent = this.vectorLayersResult .find((res) => res.values_.visible === true) .getSource() .getExtent();`
> 
> As all openlayers vec layers are visible by default, maybe there aren't any 'geoengine.vector.layer' associated with the geoengine view of your demo. In that case 'getOriginalZoom' has to detect that and use for example an empty extent in taht case:
> 
> `ol.extent.createEmpty()`

I found the error, is because any of the 'geoengine.vector.layer'  hasn't a field called 'active_on_startup' seted to True.
Normally base_geonegine needs one at least.

Many thanks!

mpascuall (approved):
Working good, checked in runboat

miquelalzanillas (approved):
Working as expected on runboat.

LGTM!

peluko00:
@dreispt Maybe it's for the dependency that not being merge?

peluko00:
@max3903 @dreispt It's ready to be merge

PR #1197, opened 2024-06-20
[15.0][ADD] fieldservice_logbook

yankinmax:
This module allows you to register and view the logbook on Field Service records.

PR #1198, opened 2024-06-26
[FIX] fieldservice_account_payment:

yankinmax:
1. `customer_id` is a required field in `fsm.location` in `fieldservice_account_analytic`
 `fieldservice_account_analytic` is installed before `fieldservice_account_payment`
 so to avoid `null value in column "customer_id" violates not-null constraint` error
 we need to depend on `fieldservice_account_analytic`
2. Wrong `inbound_payment_method_line` on bank journal which triggers this issue

https://github.com/OCA/field-service/pull/1197#issuecomment-2183143921

### DIFF
--- a/fieldservice_account_payment/static/description/index.html
+++ b/fieldservice_account_payment/static/description/index.html
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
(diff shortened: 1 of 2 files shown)


max3903 (approved):

fieldservice_account_payment/__manifest__.py
```diff
@@ -9,7 +9,11 @@
     "author": "Open Source Integrators, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/field-service",
     "depends": [
-        "fieldservice_account",
+        # customer_id is a required field in fsm.location in fieldservice_account_analytic
+        # fieldservice_account_analytic is installed before fieldservice_account_payment
+        # so to avoid `null value in column "customer_id" violates not-null constraint` error
+        # we need to depend on fieldservice_account_analytic
+        "fieldservice_account_analytic",
```
ivantodorovich: So the issue is only with unit tests, I presume?

I would try to avoid adding a possibly unwanted dependency.

Perhaps there's a different workaround, like a lazy-dependency during tests, that sets a default value through context?

yankinmax: we could try

ivantodorovich (approved):

fieldservice_account_payment/__manifest__.py
ivantodorovich: Thanks !

PR #1231, opened 2024-10-16
[18.0][MIG] fieldservice_sale

EdgarRetes:
@max3903


EdgarRetes:
@brian10048 Do you know how to prevent those warnings?

fieldservice_sale/models/sale_order.py
```diff
@@ -0,0 +1,231 @@
...
+        string="FSM Orders", compute="_compute_fsm_order_ids"
+    )
+
+    @api.depends("order_line")
+    def _compute_fsm_order_ids(self):
+        for sale in self:
+            fsm = self.env["fsm.order"].search(
+                [
+                    "|",
+                    ("sale_id", "=", sale.id),
```
brian10048: @EdgarRetes I think this line is causing the issue.  

I would try this and see if it still works and avoids the warning
```suggestion
                    ("sale_id", "in", sale.ids),
```

from https://github.com/odoo/odoo/commit/6b907f92076271f6aec5239bc0d6d9f1ab7986b1

EdgarRetes:
@brian10048 that worked, thanks. Can you check it?

brian10048:
My thoughts for improvement to make this module more cohesive with core Odoo features

We could remove the field `field_service_tracking` and combine its usage with the core Odoo field `service_tracking`

move the "Field Service Tracking" radio
Screenshot_20241107_222410

to the "Create on Order" dropdown
Screenshot_20241107_222343


This would require additional refactoring and migration scripts, but I think it is a good time to make the changes.

@max3903 would you guys be willing to work on the changes for this?

max3903:
What if someone wants to create a task and a FSM order? Same list but additional options?

max3903 (approved):

Issue #1251, opened 2024-10-25
[15.0] Error installing fieldservice_account_analytic 


iTecan:
## Module
fieldservice_account_analytic 

## Describe the bug
When you try to install the module you get the following error:
![image](https://github.com/user-attachments/assets/ed133594-b40a-45fb-bfd1-c80428c4631c)

## To Reproduce
**Affected versions**:
15.0

1. Try to install the module

**Expected behavior**
The module should be able to install without problems

If you comment the following line in the file init_hook.py you are able to install without problems:

`    cr.execute("""ALTER TABLE "fsm_location" ADD "customer_id" INT;""")
`

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)
Ubuntu 20.04
Python 3.8

max3903:
@iTecan It looks like a second attempt of the installation. Something must have gone wrong during the first and the attribute was created. Can you try removing the attribute and run the installation again?

iTecan:
We were testing installing the following modules

![image](https://github.com/user-attachments/assets/8e713afc-edf9-426a-8105-5cd272128be9)

Is a fresh install so as we understand another of the modules created the attribute first.

We test installing one by one the modules and installing directly "fieldservice_account_analytic" but in both cases the error appears.

max3903:
@iTecan Can you test and let me know if it works?

iTecan:
@max3903 we have tested the fix and works perfectly

PR #1254, opened 2024-10-31
[FIX] fieldservice_account_analytic

max3903:
#1251


iTecan:
We have check the fix and works correctly

PR #1270, opened 2024-11-27
[17.0][FIX] fieldservice: Search by name.

DorianMAG:
Remove str method in xml search filter_domain

### DIFF
--- a/fieldservice/views/fsm_order.xml
+++ b/fieldservice/views/fsm_order.xml
@@ -331,7 +331,7 @@
                 <separator />
                 <field
                     name="name"
-                    filter_domain="['|', ('name', 'ilike', self), ('name', '=like', self + '%')]"
+                    filter_domain="[('name', 'ilike', self)]"
                     string="FSM Order Name"
                 />
                 <field name="location_id" />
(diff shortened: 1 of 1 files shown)


max3903 (approved):

flotho (changes requested):
Even if I undestand the syntxa, I thinks it's useless
IMHO ('name', 'ilike', self ) should be enough
could you please give you POV @ilo @max3903  ?
Regards

flotho (approved):
as @max3903 approve I have no technical comments on this one

ivantodorovich (approved):
I agree with @flotho , that leaf seems useless and could be removed IMO.
In any case, the explicit str casting wasn't needed neither

fieldservice/views/fsm_order.xml
```diff
@@ -331,7 +331,7 @@
                 <separator />
                 <field
                     name="name"
-                    filter_domain="['|', ('name', 'ilike', self), ('name', '=like', str(self) + '%')]"
+                    filter_domain="['|', ('name', 'ilike', self), ('name', '=like', self + '%')]"
```
max3903:
```suggestion
                    filter_domain="[('name', 'ilike', self)]"
```

PR #1276, opened 2024-12-02
[17.0][FIX] fieldservice_recurring: Fix form view

jcoux:
Uses the full width to display `Field Service Orders` on `Recurring Field Service Order` form view.

Before that PR, we have:
image

With that PR, we have:
image


ivantodorovich (approved):
Thanks!

marcelsavegnago (approved):

max3903 (approved):

PR #1280, opened 2024-12-03
[17.0] [REM] fieldservice: unused `_get_stage_color` method

ivantodorovich:
The field is a non-stored related since a few versions now.
See: https://github.com/OCA/field-service/blob/9775935430d5d1919a8bc765070bfa3d7ff986af/fieldservice/models/fsm_order.py#L222

This seems to be a dead leftover code.

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -49,11 +49,6 @@ class FSMOrder(models.Model):
                 duration = delta.total_seconds() / 3600
             rec.duration = duration
 
-    @api.depends("stage_id")
-    def _get_stage_color(self):
-        """Get stage color"""
-        self.custom_color = self.stage_id.custom_color or "#FFFFFF"
-
     def _track_subtype(self, init_values):
         self.ensure_one()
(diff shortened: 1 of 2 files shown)


max3903:
@ivantodorovich It may be used on the map view.

ivantodorovich:
> @ivantodorovich It may be used on the map view.

Which map view? I searched for occurrences on this and many other oca repositories, and didn't find any other usage. But maybe I missed that one?

There are two hints that suggest it isn't used:
- The api.depends decorator, which makes no sense on a non-computed field method
- The fact that it's a private method, thus not exposed as rpc

max3903:
@ivantodorovich In `fieldservice_geoengine`

ivantodorovich:
From what I can see, it just reads this:

https://github.com/OCA/field-service/blob/2969d5cb70a75316257cc2841d3bc171568755c2/fieldservice/models/fsm_order.py#L222

yankinmax (approved):

gurneyalex (approved):

max3903 (approved):

PR #1281, opened 2024-12-03
[17.0] [IMP] fieldservice[_*]: populate `person_id` and `warehouse_id` from the related territory (if set)

ivantodorovich:
In order to achieve it in a clean and performant way, some refactors have been made to convert onchanges to computed (stored and writable) fields.

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -1,9 +1,10 @@
 # Copyright (C) 2018 Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import warnings
 from datetime import datetime, timedelta
 
-from odoo import _, api, fields, models
+from odoo import Command, _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
 
 from . import fsm_stage
@@ -112,7 +113,12 @@ class FSMOrder(models.Model):
     location_id = fields.Many2one(
         "fsm.location", string="Location", index=True, required=True
     )
-    location_directions = fields.Char()
+    location_directions = fields.Char(
+        compute="_compute_location_directions",
+        precompute=True,
+        store=True,
+        readonly=False,
+    )
     request_early = fields.Datetime(
         string="Earliest Request Date", default=datetime.now()
     )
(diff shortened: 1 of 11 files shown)


fieldservice/models/fsm_order.py
```diff
@@ -229,11 +245,77 @@ def _onchange_location_id_customer(self):
...
+            rec.description = "\n".join(
+                equipment.notes for equipment in equipments if equipment.notes
+            )
+
+    @api.depends("territory_id")
+    def _compute_person_id(self):
+        """Compute the person from the territory"""
+        for rec in self:
+            # If the person is one of the territory's workers, keep it.
+            if rec.person_id in rec.territory_id.person_ids:
```
yankinmax, line 313: Does `res.territory` model have both fields: `person_ids` and `person_id`? Just checking, because IDK

yankinmax (approved):
Thanks for the improvements

fieldservice/models/fsm_order.py
ivantodorovich: It does :)

https://github.com/OCA/field-service/blob/b6ff55995cf376cd5e64f9301f63a6a2edb2bae4/fieldservice/models/res_territory.py#L7-L11

yankinmax: Thanks for answer

max3903 (approved):

ivantodorovich:
@OCA/field-service-maintainers ok to merge?

max3903:
@ivantodorovich Yes

fieldservice_recurring/models/fsm_recurring.py
```diff
@@ -78,10 +78,28 @@ def _default_team_id(self):
...
+        store=True,
+        readonly=False,
+        index=True,
+        tracking=True,
     )
     equipment_ids = fields.Many2many("fsm.equipment")
 
+    @api.depends("location_id")
+    def _compute_person_id(self):
+        """Compute the person from the location's territory"""
```
hparfr, line 94: If I recall correctly, you may not use territory and assing person directly

ivantodorovich: The territory comes from the location (related to `location_id.territory_id`)
So the goal here is to populate the order's assigned user from the related territory's partner (if set)

Not sure if it clarifies something for you, I'm not sure I understood what you mean

hparfr: Yes but leaving person_id  blank is a feature.

The use of territory_id on location is optionnal.

I have business case where the vendor create the FSM Order and set a location (and leave person_id with no value), then (few days after) some manager assign a person to the fsm order.

ivantodorovich: That case still works, you just don't set any territory on the location, or no Assigned user on the territory.
Isn't it enough for your use case?

hparfr: If you need a "default_person_id" on territory or fsm_location, why not just create one ?
It will be more explicit about what it does.

PR #1282, opened 2024-12-04
[15.0][IMP] fieldservice_sale[_*]: Automatically cancel and regenerate FSM and picking orders with sale order changes

ppyczko:
This IMP ensures better integration between Sale, FSM and Picking Orders by:
- Automatically cancelling FSM orders when the corresponding Sale Order is cancelled.
- Allowing new FSM and Picking orders to be created if a cancelled Sale Order is reset to draft and confirmed again. When this happens, we ensure that each new FSM Order is linked to it's corresponding new Picking Order, while the previously cancelled orders remain linked to each other.

cc https://github.com/APSL 164314


peluko00 (approved):

ppyczko:
Hi @peluko00, could you review again after the new changes please? Thanks!

peluko00 (commented):
LGTM, all good. Thanks!

BernatObrador (approved):
LGTM

ppyczko:
Hi @max3903, sorry to ping directly. Could you review this when you have a moment please? Thank you!

mpascuall (approved):
LGTM, tested in runboat

fieldservice_sale_stock/models/sale_order.py
```diff
@@ -18,17 +18,18 @@ def prepare_fsm_values_for_stock_picking(self, fsm_order):
...
                     ("sale_line_id", "=", False),
+                    (
+                        "stage_id",
+                        "!=",
+                        self.env.ref("fieldservice.fsm_stage_cancelled").id,
+                    ),
                 ]
             )
-            if rec.procurement_group_id:
-                rec.procurement_group_id.fsm_order_id = fsm_order.id or False
```
ivantodorovich, line 30: Are you sure about removing these lines?

fieldservice_sale/models/sale_order.py
```diff
@@ -103,7 +103,15 @@ def _field_service_generate_sale_fsm_orders(self, new_fsm_sol):
...
         if new_fsm_sol:
             fsm_by_sale = self.env["fsm.order"].search(
-                [("sale_id", "=", self.id), ("sale_line_id", "=", False)]
+                [
+                    ("sale_id", "=", self.id),
+                    ("sale_line_id", "=", False),
+                    (
+                        "stage_id",
+                        "!=",
+                        self.env.ref("fieldservice.fsm_stage_cancelled").id,
```
max3903: @ppyczko I am pretty sure this will not work in a multi-company environment. You should use `is_closed` on the FSM order:

https://github.com/APSL/field-service/blob/15.0-imp-fieldservice_sale-cancel-order/fieldservice/models/fsm_order.py#L77

fieldservice_sale_stock/models/sale_order.py
ppyczko: Thank you for your comment @ivantodorovich! 

My thought process regarding the removal of these lines is to ensure that each new picking order generated by a sale order is associated with its own FSM order, rather than reusing the same FSM order for all pickings. My understanding is that the procurement_group_id ties together all stock operations related to the same sale order. By linking the procurement_group_id to an FSM order, we were enforcing a single FSM order across all related pickings.

Removing this line allowed each picking order to be independently linked to a distinct FSM order. This change ensures a clean one-to-one relationship between pickings and FSM orders.

fieldservice_sale/models/sale_order.py
ppyczko: Thank you for your feedback @max3903! I've applied the requested changes, let me know if there’s anything else to adjust!

fieldservice_sale_stock/models/sale_order.py
ivantodorovich: Will the `procurement.group.fsm_order_id` field be used in any other module, after this change?
If not, then maybe we need to remove it.



Additionally, I think we need to remove this related, then: https://github.com/OCA/field-service/blob/83ef6dce0e6dd72161d274c517a4a7c6d80f5d0e/fieldservice_stock/models/stock_picking.py#L9-L11

And ensure the field is populated when the picking is created instead

ppyczko: Thank you for the feedback! I’ve applied the suggested changes, it can be tested/reviewed again. Please let me know if anything else is needed.

ivantodorovich (approved):
LG

PR #1053, opened 2023-02-28
[14.0] [FIX] fieldservice_sale_recurring: Only invoice invoiceable fsm order

paradoxxxzero:
Splitting of @mourad-ehm #944 PR
Depends on #1051, #1052

> Make link between invoice and fsm order for recurring order

PR #1052, opened 2023-02-28
[14.0][IMP] fieldservice_sale: Only invoice invoiceable fsm order 

paradoxxxzero:
Splitting of @mourad-ehm #944 PR
Depends on:
- [x] #1051 

> Make link between invoice and fsm order compatible with invoice recurring order, add  _get_invoiceable_fsm_order & _get_invoiceable_fsm_order_domain

PR #1310, opened 2025-02-03
[14.0][IMP] fieldservice_sale: Only invoice invoiceable fsm order

ivantodorovich:
Reopen:
- https://github.com/OCA/field-service/pull/1052

---

It looks like we merged https://github.com/OCA/field-service/pull/1053, but not https://github.com/OCA/field-service/pull/1052, which was a direct dependency of the first one.

As a result, we've code calling `_get_invoiceable_fsm_order` (and overriding `_get_invoiceable_fsm_order_domain`), methods that don't exist in the `fieldservice_sale` module, as they were introduced by https://github.com/OCA/field-service/pull/1052

Resulting in:

```
AttributeError: 'sale.order.line' object has no attribute '_get_invoiceable_fsm_order'
```


mourad-ehm (approved):

hparfr (approved):
LGTM

PR #1312, opened 2025-02-03
[16.0][IMP] fieldservice_recurring: Expanding fields to full form width

anusriNPS:
After Migration to 16.0:
  One2many fields outside of a notebook are not expanding to the full form width by default. One should put colspan="2" for achieving such effect.

  So, included "One2many" field "fsm_order_ids" with "colspan=2" to achieve it.


brian10048:
Thank you @anusriNPS 

There is a duplicate PR for this ready for review at #1303

PR #1298, opened 2025-01-24
[15.0][ADD] fieldservice_website_sale

ppyczko:
This module integrates e-commerce with field service workflows, allowing customers to create sale orders through the e-commerce platform while selecting a preferred delivery date and time range based on their FSM route. These selections are used to determine the delivery window for the associated stock picking and field service management (FSM) orders. When an FSM order is created, it is automatically assigned to the appropriate FSM day route based on the selected delivery date. This integration streamlines operations by automating the assignment of day routes from incoming sale orders, ensuring organized and efficient scheduling.

cc https://github.com/APSL 165144

@miquelalzanillas @lbarry-apsl @mpascuall @peluko00 @javierobcn @BernatObrador please review

Depends on:
- [fieldservice_availability] #1318
- [fieldservice_sale_stock_route] #1343

PR #1318, opened 2025-02-05
[15.0][ADD] fieldservice_availability: New module fieldservice_availability

ppyczko:
This module defines blackout days (non-operational days), stress days (high-demand periods), and delivery time ranges for field service operations. It provides the necessary models to store this information, which can be used by other modules to manage scheduling, availability, and workload adjustments.

This is a technical module and does not provide functionality on its own. Extend this module to integrate availability management into field service workflows.

cc https://github.com/APSL 166637


brian10048 (approved):

mpascuall (approved):
LGTM tested in runboat

BernatObrador (approved):
LGTM

hparfr:
Hi,

Can you please explain a bit about the use case for this module ?


FYI, in fsm_recurring you can create a rrule (with the festive dates) and use it as an exclusion.

ppyczko:
> Hi,
> 
> Can you please explain a bit about the use case for this module ?
> 
> FYI, in fsm_recurring you can create a rrule (with the festive dates) and use it as an exclusion.

Hi @hparfr, thank you for your feedback. 

`fieldservice_availability` is primarily a base model for future modules that could have the need to incorporate festive days/off days or availability management. For instance:
- The `fsm_festive_day` and `fsm_delivery_time_range` models were initially part [of this PR](https://github.com/OCA/field-service/pull/1298). In `fieldservice_website_sale`, these models are used in the calendar during the checkout process, allowing customers to select a preferred delivery date and time range. The festive days block dates on the calendar to ensure customers don’t choose unavailable days.
- Another potential use case (thinking off the top of my head), could maybe be in `fieldservice_maintenance`, where we could add additional control over maintenance scheduling by blocking off dates when the company is unavailable.

Regarding the rrules from `fsm_recurring`, while `fsm_recurring` is perfect for recurring tasks, `fieldservice_availability` ensures we can handle public holidays but also a simple way to handle one-time, non-recurrent closures (like company events or non-recurrent holidays), which `fsm_recurring` doesn't address.

Let me know what you think about this.

hparfr:
I see the need to have something simplier than rrule and not refactor all the code.

Be careful with the wording about "festive". Because, in security agents business or food catering... the festive day is a working day.

I have one use case with cleaning in shops.  They need to know "specials" days like start of the sales period, etc. some time is it means they should not come (because the shop is too crowded) sometime it's the opposit: you need more cleaning.

ppyczko:
I see your point @hparfr about using the term "festive" -- it's clear that it may not work universally across all industries.

Would terms like "special days", "schedule exceptions", "service modifications" or "availability blocks" work better in this context? I'm aiming for terminology that can cover a range of use cases, where these days could either be times to avoid service or require additional services, as you suggested with the shop example.

It would be great to hear your thoughts on which term you think works best, or if you have other suggestions that might better capture this concept for different business needs. Looking forward to your input!

hparfr:
I'm not a native english speaker so I do not know the good terms. May be @brian10048 can help us on this.

May be 3 states can be good : Open / Unkown / Closed

peluko00 (approved):
LGTM, tested in runboat and code review

brian10048:
I understand the festive day is meant to represent a day where the schedule is generally unavailable.  Similar to a public holiday in Odoo Time Off. I suggest  `fsm_blackout_day` as a way to represent this.  

I had not thought of the concept requiring additional workloads.  This would be like a Stress day in Odoo Time Off.  `fsm_stress_day` ?

ppyczko:
Thank you for your suggestions @brian10048! I've implemented the changes to use `fsm_blackout_day` and `fsm_stress_day` instead of `fsm_festive_day`. Let me know if you have any other suggestions!

ppyczko:
Hi @brian10048 @max3903, is this ready to merge? Thank you in advance!

PR #1324, opened 2025-02-19
[16.0][FIX] fieldservice_geoengine: Remove widget for shape field

anusriNPS:
Widget "geo_edit_map" is not required as field type "GeoPoint" is used to identify the same. So, widget is removed for "shape" field.


HekkiMelody (approved):
The map is not shown without this change (and a warning is logged in console instead)

Code and functional review, LGTM

BernatObrador (approved):
LGTM, code review + tested in runboat.
Could be great if you port this to 17 too!

marcelsavegnago (approved):

PR #1322, opened 2025-02-11
[IMP] fieldservice_account_analytic: Invoice customer not related to any location

tafaRU:

PR #1333, opened 2025-03-05
[17.0][IMP] fieldservice_crm: Added button to create FSM orders directly from leads

BernatObrador:
This implementation adds the possibility to create FSM orders directly from the CRM form through a wizard.
It simplifies the creation of FSM orders from leads, and it will automatically create the customer's FSM location if it does not already exist.

cc https://github.com/APSL 3932

### DIFF
--- a/fieldservice_crm/README.rst
+++ b/fieldservice_crm/README.rst
@@ -92,6 +92,7 @@ Contributors
 -  `APSL <https://apsl.tech>`__:
 
    -  Antoni Marroig <amarroig@apsl.net>
+   -  Bernat Obrador <bobrador@apsl.net>
 
 Other credits
 -------------
--- a/fieldservice_crm/i18n/es.po
+++ b/fieldservice_crm/i18n/es.po
@@ -4,17 +4,16 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 16.0\n"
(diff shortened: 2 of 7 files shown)


peluko00 (approved):
LGTM, tested in runboat and code review

brian10048:
I was expecting the FSM Location to be set on the new FSM Order, but it is not populated?

image

BernatObrador:
Hi @brian10048, this issue happens when the module fieldservice_account_analytic is installed.

It seems to be caused by an onchange on the contact field that sets the location.

The thing is that this module doesn't depend on fieldservice_account_analytic, so I can't directly modify this behavior.

A possible solution would be to add a dependency on this module and apply some changes to make it work, but I'm not sure if the module author would agree with this approach.

If you try it without the fieldservice_account_analytic it will populate.

Let me know what you think!

brian10048:
@BernatObrador Ok, I was confused since the code in your PR looked correct.  I can port #1322 to 17.0 and that should solve it

BernatObrador:
@brian10048 Ok! I think with the port has to work properly.

mpascuall (approved):
LGTM, tested in runboat

ppyczko (approved):
LGTM!

brian10048 (approved):
/ocabot merge minor

PR #1334, opened 2025-03-05
[17.0] [IMP] fieldservice_account_analytic: Invoice customer not related to any location

brian10048:
Port of #1322

### DIFF
--- a/fieldservice_account_analytic/models/fsm_order.py
+++ b/fieldservice_account_analytic/models/fsm_order.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2018 - TODAY, Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class FSMOrder(models.Model):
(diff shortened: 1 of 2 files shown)


fieldservice_account_analytic/models/fsm_order.py
```diff
@@ -21,17 +21,15 @@ class FSMOrder(models.Model):
...
+    def _onchange_location_id(self):
+        self.customer_id = self.location_id.customer_id
+
     def _compute_total_cost(self):
         """To be overridden as needed from other modules"""
         for order in self:
             order.total_cost = 0.0
 
-    @api.onchange("customer_id")
-    def _onchange_customer_id_location(self):
```
BernatObrador, line 30: Maybe this onchange can be set like this instead of deleting it:
``` 
@api.onchange("customer_id")
def _onchange_customer_id(self):
      if self.customer_id.fsm_location_id:
            self.location_id = self.customer_id.fsm_location_id
      else:
            self.location_id = False
```
This approach would avoid potential buggy behaviors like:

1. Creating an FSM order.
2. Selecting a location.
3. Changing the customer to one without a location.
4. Clicking on the Location internal link.
image

However, I'm not sure if a customer can be delivered to a different location than their own.
What do you think about this approach?

BernatObrador: Another solution could be changing the context of the field to let the user freely select the location.

I guess it's just needed to delete this:
```
<field name="location_id" position="attributes">
                <attribute name="context">{"customer_id": customer_id}</attribute>
</field>
```

PR #1335, opened 2025-03-06
[17.0][FIX]  fieldservice_geoengine: Remove widget for shape field

anusriNPS:
Porting of PR [#1324](https://github.com/OCA/field-service/pull/1324)

   Widget "geo_edit_map" is not required as field type "GeoPoint" is used to identify the same. So, widget is removed from "shape" field.


HekkiMelody (approved):

hparfr (approved):
/ocabot merge minor

PR #1338, opened 2025-03-07
[FIX] fieldservice_route: clickable option for dayroute view

brian10048:
Fixes the click option on day route stages


peluko00 (approved):
LGTM

PR #1339, opened 2025-03-07
[FIX] fieldservice: display html in order report

brian10048:
Fixes the printable FSM Order report so that HTML is rendered properly for the order instructions section

Before:
image


brian10048:
this was fixed in v16 with #1162
that PR is being ported in #1340

PR #1340, opened 2025-03-07
[17.0][FW] fieldservice: multiple ports from 16.0

brian10048:
Port from 16.0 to 17.0:
- #1162
- #1163
- #1164
- #1165
- #1194
- #1202
- #1221
- #1286
- #1304
- #1317
- #1346

PR #1346, opened 2025-03-13
[16.0][IMP] fieldservice: Define @freeze_time decorator in tests and not in TestFSMOrderBase to avoid side effects

victoralmau:
Define `@freeze_time` decorator in tests and not in TestFSMOrderBase to avoid side effects

If a module has a test that extends it, a crash occurs in post tests

Related to https://github.com/OCA/field-service/pull/1340#issuecomment-2711247458

@Tecnativa


brian10048 (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1357, opened 2025-03-26
[16.0][IMP] fieldservice_geoengine: Show warning to user

anusriNPS:
Introduced "warning" to be displayed in "fsm.location" form view when value of latitude or longitude details is not available.
   
   This also helps to resolve below error which is displayed with "geoengine" view while clicking Locations from recordPanel which is observed when latitude and longitude details is not available.
   
    
Screenshot from 2025-03-10 13-35-45


HekkiMelody (approved):

HekkiMelody:
@OCA/field-service-maintainers We've been using this fix in production for months now. Will FW once merged. Thanks!

PR #1362, opened 2025-03-27
[FIX] fieldservice_sale: fix broken links in FSM order messages

brian10048:
When confirming a sale order that will generate FSM Orders, the messages logged on both sale order and FSM order have broken links

Sale Order message
Screenshot_20250327_103954

FSM Order message
Click this link but sale order is not found
Screenshot_20250327_104019


hparfr (approved):
Tested on runbot.

/ocabot merge minor

PR #1367, opened 2025-04-03
[17.0][ADD] fieldservice_base_location: New module fieldservice_base_location

BernatObrador:
This module adds the zip_id field to the fsm.location form, enabling automatic address completion based on the ZIP code entered in the location’s completion field.

To use the module, follow these steps:
1. Go to Contacts → Configuration → ZIPs.

2. Create a new ZIP entry.

3. Navigate to Field Service → Master Data → Locations.

Select an existing location or create a new one.

In the address section, you’ll now see the ZIP code autocomplete field.
Simply enter the ZIP code, and the system will automatically fill in all related address fields.

cc https://github.com/APSL 7083


ppyczko (approved):
LGTM, code review and tested in runboat!

peluko00 (approved):

lbarry-apsl (approved):
Code and runboat reviewed, LGTM!

mpascuall (approved):
LGTM!

fieldservice_address_autocomplete/models/fsm_location.py
```diff
@@ -0,0 +1,46 @@
+from odoo import models
+
+
+class FsmLocation(models.Model):
+    _inherit = "fsm.location"
+
+    def write(self, vals):
```
ivantodorovich, line 7: You don't need a  `write` override, since `fsm.location` inherits from `res.partner` the `city_id`, `state_id`, `country_id` etc are already computed whenever `zip_id` changes..

I think the only thing this module needs to do is to add the field to the view, similarly to how `base_location` is doing for the partner: https://github.com/OCA/partner-contact/blob/17.0/base_location/views/res_partner_view.xml, with possibly also its `_fields_view_get` override if needed, to accomplish the same UX than the partner view

ivantodorovich (changes requested):
IMO this module should be named `fieldservice_base_location` or `fieldservice_partner_location`, and be auto-installed when `base_location` is installed

fieldservice_address_autocomplete/models/fsm_location.py
BernatObrador: Hi @ivantodorovich, thanks for your feedback.

The thing is that it really needs the overwritten `write` method, otherwise it ends up throwing errors due to mismatches in partner fields.

For example, when you change the ZIP code or use the autocomplete in the field service location, saving will throw an error because the partner data does not match the ZIP.

The only solution I found was to overwrite the `write` method.

Regarding the module name, `fieldservice_base_location` is fine for me.

Let me know your thoughts!

fieldservice_base_location/tests/test_address_autocomplete.py
```diff
@@ -0,0 +1,102 @@
...
+                "zip": "07001",
+                "zip_id": cls.zip_record.id,
+                "owner_id": cls.res_partner.id,
+            }
+        )
+
+    @classmethod
+    def _request_handler(cls, s, r, /, **kw):
+        """Don't block external requests."""
+        return cls._super_send(s, r, **kw)
```
ivantodorovich, line 52: What's this for?

```suggestion
```

fieldservice_address_autocomplete/models/fsm_location.py
ivantodorovich: Thanks @BernatObrador 

Is this covered by the unit tests you wrote?
Meaning, would they fail without the write override?

If we keep it, then, please write a detailed comment explaining the error you get and how this solves it.

```py
def write(self, vals):
    # OVERRIDE to ...
    # ...
    # ...
```

fieldservice_base_location/tests/test_address_autocomplete.py
BernatObrador: This is needed cause it tries to geolocate it, so an external request is sent.

fieldservice_address_autocomplete/models/fsm_location.py
BernatObrador: Yes it's.

The test_partner_fields_synced, covers it, as it ensures that all fields are synced.

ivantodorovich (approved):
/ocabot merge nobump

PR #1382, opened 2025-04-22
[ADD] fieldservice_check_worker_availability: New module fieldservice_check_worker_availability

mpascuall:
This module raises an error when attempting to create a new FSM Order if 
the selected start date falls within a worker's non-working or leave period.

It also removes the group restriction on the Time Off smart button 
so it becomes visible without enabling debug mode.

cc https://github.com/APSL 8363
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00  @BernatObrador @ppyczko please review


peluko00 (approved):

fieldservice_check_worker_availability/models/fsm_order.py
```diff
@@ -0,0 +1,90 @@
...
+                        "the selected date and time: %(date)s."
+                    )
+                    % {
+                        "name": person.name,
+                        "date": scheduled_date_dt_utc.strftime("%Y-%m-%d %H:%M"),
+                    }
+                )
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
BernatObrador: Create and write, can be replaced with @api.constrains("scheduled_date_start", "person_id")?

it will end up in one function like this:
```python
@api.constrains("scheduled_date_start", "person_id")
def check_worker_availability(self):
   for order in self:
      if order.scheduled_date_start and order.person_id:
         self._check_worker_availability(
            order.scheduled_date_start, 
            order.person_id
          )
```

BernatObrador (changes requested):

fieldservice_check_worker_availability/models/fsm_order.py
miquelpascual: Done, thanks! Could you review again please?

BernatObrador (approved):
LGTM!

ppyczko (changes requested):
While testing this module, I noticed that it's still possible to create an FSM order that overlaps with a time off entry. The current validation only checks the Scheduled Start (ETA), but doesn't account for the entire scheduled duration of the order.

For example, consider this time off entry:

imagen

If I then create an FSM order starting at 8:45h with a scheduled duration of 4 hours (ending at 12:45h), no error is raised—even though the FSM order overlaps the entire time off period (which goes from 9:00h to 11:00h).

imagen

The same issue occurs even if the order only partially overlaps the time off. For example, in another test, the time off entry again spans from 9:00h to 11:00h, and the FSM order overlaps it by 1 hour and 45 minutes.

imagen

miquelpascual:
> While testing this module, I noticed that it's still possible to create an FSM order that overlaps with a time off entry. The current validation only checks the Scheduled Start (ETA), but doesn't account for the entire scheduled duration of the order.
> 
> For example, consider this time off entry:
> 
> imagen
> 
> If I then create an FSM order starting at 8:45h with a scheduled duration of 4 hours (ending at 12:45h), no error is raised—even though the FSM order overlaps the entire time off period (which goes from 9:00h to 11:00h).
> 
> imagen
> 
> The same issue occurs even if the order only partially overlaps the time off. For example, in another test, the time off entry again spans from 9:00h to 11:00h, and the FSM order overlaps it by 1 hour and 45 minutes.
> 
> imagen

Hi! Many thanks, could you review it again please?

fieldservice_check_worker_availability/models/fsm_order.py
```diff
@@ -0,0 +1,81 @@
...
+                    _(
+                        "%(name)s has a registered leave on "
+                        "this date %(date)s.\nReason: %(reason)s"
+                    )
+                    % {
+                        "name": person.name,
+                        "date": scheduled_date_dt_utc.strftime("%Y-%m-%d %H:%M"),
+                        "reason": overlapping_leave.name,
+                    }
+                )
```
ppyczko: Tested functionality and it works great. As a last suggestion, could you please modify this ValidationError message to be more accurate? 

Right now, it only shows the start time of the FSM order, which can be misleading in cases where the order overlaps partially with a leave (e.g., FSM order scheduled from 13:30h to 15:30h, leave from 14:00h to 15:00h). The message that shows is:

imagen

But the leave is registered from 14:00h to 15:00h.

To make the overlap clearer, maybe we could include both the start and end time of the order and the leave. Something like this, for example:
```suggestion
                raise ValidationError(
                    _(
                         "%(name)s has a registered leave from %(leave_start)s to %(leave_end)s, "
                         "which overlaps with the scheduled time of this order "
                         "(from %(order_start)s to %(order_end)s).\nReason: %(reason)s"
                    )
                    % {
                        "name": person.name,
                        "leave_start": overlapping_leave.date_from.strftime("%Y-%m-%d %H:%M"),
                        "leave_end": overlapping_leave.date_to.strftime("%Y-%m-%d %H:%M"),
                        "order_start": scheduled_date_dt_utc.strftime("%Y-%m-%d %H:%M"),
                        "order_end": scheduled_date_end_dt_utc.strftime("%Y-%m-%d %H:%M"),
                        "reason": overlapping_leave.name,
                    }
                )
```

miquelpascual: Good idea, could you review it now?

ppyczko (approved):
LGTM! 🚀

PR #1383, opened 2025-04-25
[17.0][IMP] fieldservice: Improve partner location creation with new partner

peluko00:
In the creation of partner location the field `partner_id` is readonly in form view and you can't create a new partner from that view.
This improvement remove the required  `owner_id` field and put the required on `partner_id` that can create partner from the same screen without changing the application or the view.

cc https://github.com/APSL 1811

### DIFF
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -25,7 +25,6 @@ class FSMLocation(models.Model):
     owner_id = fields.Many2one(
         "res.partner",
         string="Related Owner",
-        required=True,
         ondelete="restrict",
         auto_join=True,
     )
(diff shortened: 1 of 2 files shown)


max3903 (commented):
I'd like to keep the partner_id field hidden by default. 
Same as in the user with the related partner field.

peluko00:
> I'd like to keep the partner_id field hidden by default. Same as in the user with the related partner field.

Hi @max3903, but this kind of flow can't create new partner from the same view and is very tedious job. What do you think about it?

max3903:
@peluko00 Odoo is doing it with user and partner for 12+ millions users :)

PR #1387, opened 2025-05-06
[17.0][IMP] fieldservice: Link teams with workers

peluko00:
Link teams with workers

cc https://github.com/APSL 9530
@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall @BernatObrador @ppyczko please review

PR #1388, opened 2025-05-07
[FIX] fieldservice_vehicle: Fix onchange person_id

peluko00:
When person_id change the team is not setting.
When override the onchange method you need to call super for make the inherit functions from the basic module working. See the code below:
https://github.com/OCA/field-service/blob/17.0/fieldservice/models/fsm_order.py#L400

#1387 
cc https://github.com/APSL 9530
@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall @BernatObrador @ppyczko please review

### DIFF
--- a/fieldservice_vehicle/models/fsm_order.py
+++ b/fieldservice_vehicle/models/fsm_order.py
@@ -29,3 +29,4 @@ class FSMOrder(models.Model):
             and self.person_id.vehicle_id.id
             or False
         )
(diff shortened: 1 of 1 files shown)


BernatObrador (approved):
LGTM

mpascuall (approved):
LGTM!

max3903 (approved):

PR #1353, opened 2025-03-19
[15.0][IMP] fieldservice_activity: Use @freeze_time for time-sensitive tests

ppyczko:
When trying to merge #1348 we got an error on this [line](https://github.com/OCA/field-service/blob/15.0/fieldservice_activity/tests/test_fsm_activity.py#L82) due to a one second difference:

`2025-03-17 15:43:02,000 410 ERROR odoo odoo.addons.fieldservice_activity.tests.test_fsm_activity: FAIL: TestFSMActivity.test_fsm_activity Traceback (most recent call last): File "/__w/field-service/field-service/fieldservice_activity/tests/test_fsm_activity.py", line 82, in test_fsm_activity self.assertEqual( AssertionError: datetime.datetime(2025, 3, 17, 15, 43, 1) != datetime.datetime(2025, 3, 17, 15, 43, 2)`

This PR prevents such failures by adding the `@freeze_time` decorator to `test_fsm_activity` method.

PR #1392, opened 2025-05-14
[15.0][FIX] fieldservice_sale_stock_route: Use @freeze_time for time-sensitive tests

ppyczko:
When trying to merge #1298 we got an error due to a one second difference:

`Traceback (most recent call last):
  File "/__w/field-service/field-service/fieldservice_sale_stock_route/tests/test_fieldservice_sale_stock_route.py", line 60, in test_commitment_dates_on_confirmation
    self.assertEqual(
AssertionError: datetime.datetime(2025, 5, 15, 15, 22, 14) != datetime.datetime(2025, 5, 15, 15, 22, 13) : Commitment date end should match the next route day after confirmation.`

This PR prevents such failures by adding the @freeze_time decorator to all time sensitive tests.

cc https://github.com/APSL 9930

### DIFF
--- a/fieldservice_sale_stock_route/tests/test_fieldservice_sale_stock_route.py
+++ b/fieldservice_sale_stock_route/tests/test_fieldservice_sale_stock_route.py
@@ -3,6 +3,8 @@
 
 from datetime import datetime, timedelta
 
+from freezegun import freeze_time
+
 from odoo.exceptions import ValidationError
 from odoo.tests.common import TransactionCase
 
@@ -44,6 +46,7 @@ class TestFieldServiceSaleStockRoute(TransactionCase):
             }
         )
 
+    @freeze_time("2025-05-15")
     def test_commitment_dates_on_confirmation(self):
         """
         Test that commitment_date and commitment_date_end are correctly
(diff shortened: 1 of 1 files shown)


mpascuall (approved):
LGTM!

lbarry-apsl (approved):
LGTM!

fieldservice_sale_stock_route/tests/test_fieldservice_sale_stock_route.py
```diff
@@ -44,6 +46,7 @@ def setUpClass(cls):
             }
         )
 
+    @freeze_time("2025-05-15")
```
peluko00, line 49: Maybe a good solution is compare only the day and not the time in every case. What do you think?

ppyczko: Thank you for the suggestion. In my opinion, using @freeze_time instead of comparing just the date values is the better approach here, as it avoids cluttering the test with verbose assertions. Since we're consistently working with datetime objects throughout the codebase, reducing them to just dates in tests would introduce inconsistency and reduce readability.

Field service operations are generally time-sensitive, so I also believe it’s appropriate, and necessary, to compare full datetime values, not just the date.

Moreover, this approach has already been applied consistently in other PRs, such as #1353, so I believe sticking with freeze_time is the right and proven solution here.

PR #1411, opened 2025-06-04
[16.0][FIX][fieldservice_reccuring]: Fix error in search view

anusriNPS:
Fixed error seen in search view under Operations > Recurring Orders for field name
"FSM Recurring Order Name".

Error shown as below:

image

### DIFF
--- a/fieldservice_recurring/views/fsm_recurring.xml
+++ b/fieldservice_recurring/views/fsm_recurring.xml
@@ -114,11 +114,7 @@
         <field name="model">fsm.recurring</field>
         <field name="arch" type="xml">
             <search string="Recurring Orders">
-                <field
-                    name="name"
-                    filter_domain="['|', ('name', 'ilike', self), ('name', '=like', str(self) + '%')]"
-                    string="FSM Recurring Order Name"
-                />
+                <field name="name" string="FSM Recurring Order Name" />
                 <field name="person_id" />
                 <field name="location_id" />
                 <field name="team_id" groups="fieldservice.group_fsm_team" />


fieldservice_recurring/views/fsm_recurring.xml
```diff
@@ -116,7 +116,7 @@
             <search string="Recurring Orders">
                 <field
                     name="name"
-                    filter_domain="['|', ('name', 'ilike', self), ('name', '=like', str(self) + '%')]"
+                    filter_domain="['|', ('name', 'ilike', self), ('name', '=like', self + '%')]"
```
HekkiMelody: **question**: what is the point of having such a complicated `filter_domain`? I'm guessing this is probably leftover from previous versions.

Unless I'm missing something, any order matched by `('name', '=like', self + '%')` would also be matched by `('name', 'ilike', self)`

So maybe we could simply have `<field name="name"/>` without any custom domain? Wouldn't that effectively get us the same result?

HekkiMelody (approved):
Current PR is the smallest possible change that fixes the problem, so approving anyway.

However, we can probably simplify further.

fieldservice_recurring/views/fsm_recurring.xml
anusriNPS: Yes, even I was thinking the same as it does not have any value. I left it same to preserve it as it was not creating any issue.

Even without filter domain,  it fetches same results. So, updated without filter_domain

HekkiMelody (approved):

ivantodorovich (approved):
/ocabot merge patch

PR #1416, opened 2025-06-10
[15.0][ADD] fieldservice_sale_note: New module fieldservice_sale_note

peluko00:
This module extends the integration between Sale Orders and Field Service Orders by automatically transferring the Internal Note from the related Sale Order into the Instructions field of the corresponding Field Service Order. 

cc https://github.com/APSL HT02090


mpascuall (approved):
LGTM!

ppyczko (approved):
LGTM!

fieldservice_sale_note/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module extends the integration between Sale Orders and Field Service Orders by automatically transferring the Internal Note from the related Sale Order into the Resolution field of the corresponding Field Service Order.
```
ivantodorovich: I think you should update this to mention the fact that the Sale order's "Internal Note" field is also added by this module.

Or perhaps even better:
- Extract the Sale Order's Internal note field addition to a specific `sale_order_internal_note` module in OCA/sale-workflow


~However, this may conflict with core's enterprise `sale_suscription` module: https://github.com/odoo/enterprise/blob/d6fa3be5907a3b7afab2af39f60aa234bb09159f/sale_subscription/models/sale_order.py#L126-L127~

~So perhaps the name should be different (e.g.: `note` instead of `internal_note`)~

My bad: I was looking at the wrong Odoo version

ivantodorovich (changes requested):

fieldservice_sale_note/readme/DESCRIPTION.md
peluko00: Fixed! Can you review again please @ivantodorovich ?

fieldservice_sale_note/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+Extract the Sale Order's Internal note field addition to a specific sale_order_internal_note module in OCA/sale-workflow
```
ivantodorovich: Actually I meant this as an action to do, not as the readme content 😓 

But if your not going for it, then I guess you could write this in the ROADMAP section

peluko00: I think that's it's ready now for merge with this README content

ivantodorovich: I think we're not on the same page, no.

Can you move this to a `ROADMAP.md` file:
```
Extract the Sale Order's Internal note field addition to a specific sale_order_internal_note module in OCA/sale-workflow
```


And on this `DESCRIPTION.md`, please keep a description of what the module is doing
For example:
```
Allows to register an internal note on Sales Orders, and pass it through to the FSM Orders
```

peluko00: Done! Can you review again please @ivantodorovich ?

ivantodorovich (approved):
Thanks!

/ocabot merge nobump

PR #1417, opened 2025-06-19
[17.0][IMP] fieldservice: Add timezone-aware date filters for FSM orders

BernatObrador:
This commit introduces date_today_order_tz computed field in the `fsm.order` model
to enable accurate filtering of orders based on the user's timezone.

These fields allow view filters (e.g., "Today Orders") to function
correctly regardless of timezone, addressing the limitation where
UTC-stored datetimes cause inconsistencies in local time filtering.

cc https://github.com/APSL 13151

### DIFF
--- a/fieldservice/models/fsm_order.py
+++ b/fieldservice/models/fsm_order.py
@@ -3,6 +3,8 @@
 
 from datetime import datetime, timedelta
 
+import pytz
+
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
 from odoo.tools import format_date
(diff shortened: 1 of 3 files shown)


BernatObrador:
Hi @brian10048, @max3903!

Let me know what you think about this approach. This change is really important to ensure that date-based filters (like "Today" or "This Week") behave correctly based on the user's timezone.

Without this, filters can include or exclude records incorrectly—especially around midnight—since datetimes are stored in UTC. The computed fields allow us to define stored, timezone-aware filters directly in views.

Happy to refine this if needed!

ppyczko (approved):
LGTM, tested in runboat

mpascuall (approved):
LGTM!

max3903 (approved):

PR #1429, opened 2025-08-04
[18.0][MIG] fieldservice_portal: Migration to 18.0

peluko00:
Module migrated to version 18.0

cc https://github.com/APSL 17756

- [Issue] https://github.com/OCA/field-service/issues/1220


mpascuall (approved):
LGTM

fieldservice_portal/controllers/fsm_order_portal.py
```diff
@@ -15,13 +15,12 @@ class CustomerPortal(CustomerPortal):
...
-                if request.env["fsm.order"].check_access_rights(
-                    "read", raise_exception=False
+            try:
+                fsm_order_count = request.env["fsm.order"].search_count(
+                    self._prepare_fsm_orders_domain()
                 )
-                else 0
-            )
+            except AccessError:
+                fsm_order_count = 0
```
ivantodorovich: This is not the correct way to solve.
The counter should be shown properly, so possibly use sudo() for the search_count

ivantodorovich (changes requested):
Overall seems good. Thanks

However, there are two issues (possibly the same).

The CI is failing due to AccessError, so there's somethign to fix there. Could you take a look?

fieldservice_portal/controllers/fsm_order_portal.py
peluko00: Solve! Can you review again please @ivantodorovich ?

ivantodorovich:
Thanks!

The CI is still failing though:

```
errors that caused failure (1):
2025-08-19 07:51:08,383 267 WARNING odoo odoo.http: You are not allowed to access 'Field Service Stage' (fsm.stage) records.

This operation is allowed for the following groups:
	- Field Service/Manager
	- Field Service/User (only own documents)
	- User types/Portal

Contact your administrator to request access if necessary. 
```

peluko00:
> Thanks!
> 
> The CI is still failing though:
> 
> ```
> errors that caused failure (1):
> 2025-08-19 07:51:08,383 267 WARNING odoo odoo.http: You are not allowed to access 'Field Service Stage' (fsm.stage) records.
> 
> This operation is allowed for the following groups:
> 	- Field Service/Manager
> 	- Field Service/User (only own documents)
> 	- User types/Portal
> 
> Contact your administrator to request access if necessary. 
> ```

I don't know about the problem because the user have rights about it. Can you have any idea @ivantodorovich ?

ivantodorovich:
I can't tell by reading the code.
I'm suspicious about the part that tests the flow with the "demo" user instead of "portal".

I suggest to try to reproduce locally and debug it.

ivantodorovich (commented):
Quite possibly is this:

```
        # Check Forbidden response from API
        self.assertEqual(response.status_code, 403)
```

If so, we just need to mute the logger there.
The test is correct, but the fact that the warning is emitted is not --> which causes the CI to fail

peluko00:
> Quite possibly is this:
> 
> ```
>         # Check Forbidden response from API
>         self.assertEqual(response.status_code, 403)
> ```
> 
> If so, we just need to mute the logger there. The test is correct, but the fact that the warning is emitted is not --> which causes the CI to fail

Fixed, thank you! It's ready for merge @ivantodorovich

ivantodorovich (approved):
/ocabot merge nobump

Thanks!

PR #1433, opened 2025-08-07
[18.0] [IMP] fieldservice: use computed fields instead of onchanges

ivantodorovich:
Modernize these bits


imlopes (approved):

vvrossem (approved):
LGTM, thank you

PR #1440, opened 2025-08-18
[18.0][ADD] fieldservice_account_vendor_bill

imlopes:


fieldservice_account_vendor_bill/models/fsm_order.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import fields, models
+
+
+class FSMOrder(models.Model):
+    _inherit = "fsm.order"
+
+    vendor_bill_ids = fields.Many2many(
+        "account.move",
+        string="Vendor bills",
```
vvrossem, line 12:
```suggestion
```

vvrossem (approved):
LG, but why not adding this to `fieldservice_account` instead of adding a new module?

PR #1441, opened 2025-08-19
[18.0][IMP] fieldservice_sale: sale.order::commitment_date propagation

vvrossem:
Propagates `sale.order::commitment_date` to `fsm.order::scheduled_date_start` (on create and write)

### DIFF
--- a/fieldservice_sale/models/sale_order.py
+++ b/fieldservice_sale/models/sale_order.py
@@ -89,7 +89,7 @@ class SaleOrder(models.Model):
             "location_id": self.fsm_location_id.id,
             "location_directions": self.fsm_location_id.direction,
             "request_early": self.expected_date,
-            "scheduled_date_start": self.expected_date,
+            "scheduled_date_start": self.commitment_date or self.expected_date,
             "todo": note,
             "category_ids": [(6, 0, categories.ids)],
             "scheduled_duration": hours,
@@ -229,3 +229,14 @@ class SaleOrder(models.Model):
         else:
             action = {"type": "ir.actions.act_window_close"}
         return action
+
+    def write(self, values):
+        res = super().write(values)
+        if "commitment_date" in values:
+            scheduled_date_start = values["commitment_date"]
+            for order in self:
+                for fsm_order in order.fsm_order_ids:
+                    fsm_order.scheduled_date_start = (
+                        scheduled_date_start or order.expected_date
+                    )
+        return res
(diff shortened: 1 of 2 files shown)


fieldservice_sale/models/sale_order.py
```diff
@@ -229,3 +229,18 @@ def action_view_fsm_order(self):
...
+            # So we skip the commitment_date propagation if the sale order is in draft.
+            # That means that the FSM Orders are not created yet.
+            # We also skip if the sale order is cancelled as it makes no sense.
+            if (
+                order.state not in ("draft", "cancel")
+                and vals.get("commitment_date") is not None
+            ):
+                for fsm_order in order.fsm_order_ids:
+                    fsm_order.scheduled_date_start = vals["commitment_date"]
+        return res
```
ivantodorovich: I think here we should align to what Odoo does in `sale_stock`: https://github.com/odoo/odoo/blob/6194ca0042f967308178699837f01d42f46dea52/addons/sale_stock/models/sale_order.py#L143-L148

Meaning:
- Sync even if the sale order is confirmed
- Fallback to order.expected_date if there's no commitment_date

fieldservice_sale/models/sale_order.py
```diff
@@ -89,7 +89,7 @@ def _prepare_fsm_values(self, **kwargs):
             "location_id": self.fsm_location_id.id,
             "location_directions": self.fsm_location_id.direction,
             "request_early": self.expected_date,
-            "scheduled_date_start": self.expected_date,
+            "scheduled_date_start": self.commitment_date,
```
ivantodorovich: Just like with my other comment, I'd align to core here: https://github.com/odoo/odoo/blob/66c5d10833af8946ab5d3f887086f9ba08d36ab3/addons/sale_stock/models/sale_order_line.py#L260

```suggestion
            "scheduled_date_start": self.commitment_date or self.expected_date,
```

ivantodorovich (approved):
Thanks Vincent! Looks great

ivantodorovich:
The ci disagrees, though.
Looks like there's a unit test to adapt:

```

errors that caused failure (4):
2025-08-19 13:35:30,054 266 ERROR odoo odoo.addons.fieldservice_sale.tests.test_fsm_sale_order: FAIL: TestFSMSaleOrder.test_sale_order_6
  File "/__w/field-service/field-service/fieldservice_sale/tests/test_fsm_sale_order.py", line 566, in test_sale_order_6
    self.assertFalse(
AssertionError: datetime.datetime(2025, 8, 19, 13, 35, 29) is not false : FSM Sale: FSM Orders should have empty scheduled start date
```

yankinmax (approved):

PR #1226, opened 2024-10-15
[18.0][MIG] fieldservice_account_analytic

EdgarRetes:
@max3903

PR #1442, opened 2025-08-19
[18.0][MIG] fieldservice_isp_account: Migration to 18.0

vvrossem:
Depends on

* https://github.com/OCA/field-service/pull/1226

some other modules unit tests are failing because `fieldservice_account_analytic` introduces required field `fsm.location::customer_id`.
Example: `fieldservice_crm` (which doesn't depend on `fieldservice_account_analytic`) fails to create a `fsm.location`.


ivantodorovich:
In case others are looking for this module, we've decided to abandon it due to the issues mentioned above and in the `fieldservice_account_analytic` dependency. Instead, we are replacing it with:
- https://github.com/OCA/field-service/pull/1450

For our simple use case of registering timesheets on work orders it's enough.

PR #1445, opened 2025-09-02
[16.0][IMP] fieldservice_geoengine: Legends based on attribute_field_id

anusriNPS:
When "attribute_field_id" of geoengine data is configured
with other than "stage_name" field, legends are not displayed based on fields configured.
   Replacing legend is required only for "custom_color" as legend
needs to be displayed based on "stage_name", if not legend points to "custom_color" as value in geoengine map view.


HekkiMelody (approved):
Code review, LGTM

ivantodorovich (approved):

PR #1446, opened 2025-09-09
[IMP] fieldservice_sale_order_create_wizard: extract sale order values into helper method

ppyczko:
Move sale order values from `action_create_sale_order` to `_prepare_sale_order_vals` for cleaner code and easier extension.


BernatObrador (approved):
LGTM!

PR #1448, opened 2025-09-11
[18.0][FIX] fsm.order.todo displays HTML tags in Qweb report

vvrossem:
#### Before
Screenshot from 2025-09-11 09-16-13

#### After
image


ivantodorovich (approved):

yankinmax (approved):
Just wonder if we need to provide a migration script to handle existing records

PR #1449, opened 2025-09-17
[18.0][FIX] fieldservice: fsm.order::onchange_template_id must be public

vvrossem:
`_onchange_template_id` is not triggered **on the UI**, whereas `onchange_template_id` is.


vvrossem:
Not the actual issue.

PR #1455, opened 2025-10-10
[17.0] [FIX] fieldservice_recurring: order creation vals

brian10048:
This change prevents calling `fsm_order._onchange_template_id()` right after creating a new `fsm_order` from the `fsm_recurring`

Previously, the `scheduled_duration` and `team_id` fields were set to the order template value in the onchange method, which changes the value initially returned by the `fsm_recurring._prepare_order_values()` method.


ivantodorovich (approved):
Change looks good.

CI seems to be failing for an unrelated reason, though, possibly a timeout?
Maybe restart the github action and see if it resoves

brian10048:
@ivantodorovich I updated this one with some tests

PR #1459, opened 2025-10-24
[17.0][IMP] fieldservice_current_location: add configurable and unique Nominatim user agent + mock tests

BernatObrador:
Generate a persistent and unique User-Agent per instance and store it in ir.config_parameter to comply with Nominatim's usage policy.
Add tests mocking Nominatim and RateLimiter to avoid external HTTP requests.

This change prevents instances from being blocked by Nominatim, as happened in [this PR run](https://github.com/OCA/field-service/actions/runs/18770594799/job/53554248813?pr=1458) , which caused test failures due to real network calls.

It also ensures compliance with Nominatim’s usage requirements by setting a proper User-Agent for all requests.


BernatObrador:
@ivantodorovich Could you review please? Thanks for your time!

ivantodorovich (approved):
I'm not experienced in this module, but the solution looks technically good 👍🏻

brian10048 (approved):
/ocabot merge minor

PR #1460, opened 2025-10-28
[18.0] [REF] fieldservice: proper fsm.location hierarchy implementation

ivantodorovich:
This commit implements a proper fsm.location hierarchy with `parent_id` and parent_store / parent_path.

It makes possible to use hierarchy domain operators (`child_of`, `parent_of`) and allows for more efficient queries.


fieldservice/models/fsm_location.py
```diff
@@ -147,177 +154,121 @@ def _onchange_district_id(self):
...
+        else:
+            action["domain"] = domain
+        return action
 
-    def _compute_contact_ids(self):
+    def _compute_contact_count(self):
+        if not self.ids:  # pragma: no cover
+            self.contact_count = 0
+            return
+        count_by_location = dict[Self, int](
```
yankinmax, line 181: this is something new to me and I can't find source code in `odoo.api`

yankinmax (approved):
Thanks @ivantodorovich for modern way refactor with correct counting and `child of` usage. You rock!

fieldservice/models/fsm_location.py
```diff
@@ -54,72 +56,77 @@ class FSMLocation(models.Model):
...
+        self.contact_id = self.parent_id.contact_id
+        self.direction = self.parent_id.direction
+        self.street = self.parent_id.street
+        self.street2 = self.parent_id.street2
+        self.city = self.parent_id.city
+        self.zip = self.parent_id.zip
+        self.state_id = self.parent_id.state_id
+        self.country_id = self.parent_id.country_id
+        self.tz = self.parent_id.tz
+        self.territory_id = self.parent_id.territory_id
```
SilvioC2C, line 129: Could it be worth it to add `inverse="_inverse_parent_id"` in the field's attributes to trigger this change? Or do we want this function to be triggered in the UI only?

SilvioC2C (approved):
LGTM, one minor, non-blocking doubt

fieldservice/models/fsm_location.py
ivantodorovich: TBH I think we need to get rid of all (or most) onchanges and move to stored computed writeable fields, but that's for another PR

PR #1467, opened 2025-11-10
[17.0][FIX] fieldservice: Use date_stop on calendar view

BernatObrador:
Avoids fetching all service orders when opening the calendar view, which caused unnecessary queries and performance issues. Now only records within the selected time period are retrieved.

For example, viewing the calendar in week mode:

### Using the date_stop
it will only fetch the registers of the selected week:
image

the domain used to fetch the records:

image

### Not using date_stop
All service orders created before the last day shown in the calendar view are fetched, leading to excessive queries and degraded performance:
image

the domain used to fetch the records:

image

### DIFF
--- a/fieldservice/views/fsm_order.xml
+++ b/fieldservice/views/fsm_order.xml
@@ -461,6 +461,7 @@
             <calendar
                 string="FSM Orders"
                 date_start="scheduled_date_start"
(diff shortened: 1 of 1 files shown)


brian10048 (approved):

ivantodorovich (approved):
/ocabot merge patch

PR #1469, opened 2025-11-25
[18.0] [IMP] fieldservice: make some notes fields Html

ivantodorovich:


vvrossem (approved):
LG :racing_car:

fieldservice/models/fsm_order.py
```diff
@@ -282,7 +282,7 @@ def _compute_description(self):
         for rec in self:
             if rec.description:
                 continue
-            rec.description = "\n".join(
+            rec.description = "<separator />".join(
```
yankinmax: Shouldn't it be declared as `Markup` for Html field type?
`Markup("<separator />").join()`

yankinmax (approved):

fieldservice/models/fsm_order.py
ivantodorovich: I'm not sure, I'll change just in case 👍🏻

PR #1471, opened 2025-11-26
[18.0] [IMP] fieldservice: drop is_button field

ivantodorovich:
This field was only used to ensure an order can only be moved to a Completed stage through the "Complete" button (action_complete).

Instead, rely on a context flag to bypass the validation.

### DIFF
--- a/fieldservice/tests/test_fsm_order.py
+++ b/fieldservice/tests/test_fsm_order.py
@@ -7,6 +7,7 @@ from odoo import fields
 from odoo.exceptions import UserError, ValidationError
 from odoo.tests import Form
 from odoo.tests.common import TransactionCase
+from odoo.tools import mute_logger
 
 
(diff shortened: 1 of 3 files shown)


fieldservice/tests/test_fsm_order.py
```diff
@@ -264,23 +272,22 @@ def test_fsm_order(self):
...
-                {
-                    "location_id": self.test_location.id,
-                    "stage_id": self.stage1.id,
-                }
-            )
         order.can_unlink()
         order.unlink()
 
+    def test_order_move_to_completed(self):
+        """Test move to compelted
```
vvrossem:
```suggestion
        """Test move to completed
```

vvrossem (approved):

yankinmax (approved):
Thanks @ivantodorovich , a nice improvement

PR #1472, opened 2025-11-26
[18.0] [IMP] fieldservice: add Customer Signature field

ivantodorovich:
Makes it possible to ask for a Customer Signature when the order is in a given stage.


fieldservice/views/fsm_order.xml
```diff
@@ -17,6 +17,21 @@
...
+                        class="oe_highlight"
+                        invisible="not require_signature or signature"
+                    />
+                    <button
+                        string="Sign"
+                        id="action_sign_2"
+                        type="action"
+                        name="%(fsm_order_sign_wizard_action)d"
+                        invisible="not require_signature or not signature"
+                    />
```
vvrossem, line 34: I don't understand why there are 2 "Sign" buttons

ivantodorovich: One has `class="oe_highlight"` and the other doesn't (shown when there's already a signature)

vvrossem: my bad, I've misread the invisible condition :flushed:

yankinmax (approved):

vvrossem (approved):

PR #1484, opened 2025-12-15
[18.0] [FIX] fieldservice_repair: fix context propagation when creating repair orders

ivantodorovich:


fieldservice_repair/models/fsm_order.py
```diff
@@ -80,9 +81,13 @@ def _create_repair_orders(self):
...
                     )
                 repair_order_vals = rec._prepare_repair_order_vals(equipment)
-                repair_order = self.env["repair.order"].create(repair_order_vals)
-                created_repair_orders += repair_order
-        return created_repair_orders
+                repair_order_vals_list.append(repair_order_vals)
+        # pylint: disable=context-overridden
+        return (
+            self.env["repair.order"]
+            .with_context(clean_context(self.env.context))
```
yankinmax, line 88: TIL: `clean_context`.
But, why do we need here:
> remove each entry with its key starting with ``default_``

yankinmax (approved):

fieldservice_repair/models/fsm_order.py
vvrossem: If I understood correctly, we want to remote `default_priority` key because its value (e.g. `'2`') is not compatible with the ones of repair.order.priority: https://github.com/odoo/odoo/blob/18.0/addons/repair/models/repair.py#L54

vvrossem (approved):

fieldservice_repair/models/fsm_order.py
ivantodorovich: That's right!

yankinmax: Thanks for explanation

PR #1493, opened 2026-01-05
[18.0][FIX] fieldservice_project: fsm_location go to projects

norlinhenrik:
I have removed a domain that is not working.


brian10048 (approved):
Tested on runboat and working as expected

PR #1496, opened 2026-01-16
[19.0][MIG] fieldservice: Migration to 19.0

hoattk-komit:


Vang-NguyenPhu (approved):
LGTM

huan-tq (approved):
LGTM! Tested in our environment!
image

fieldservice/test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-base_territory @ git+https://github.com/OCA/field-service.git@refs/pull/1495/head#subdirectory=base_territory
```
cuongnmtm: This change should be in a separate commit  https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29

cuongnmtm (changes requested):

brian10048:
@hoattk-komit the dependent module is now merged.  Can you remove the test requirements commit?

hoattk-komit:
The file `test-requirements.txt` was removed.

fieldservice/views/fsm_location.xml
```diff
@@ -105,6 +106,7 @@
                             <field
                                 name="partner_id"
                                 groups="base.group_no_one"
+                                required="0"
```
cuongnmtm, line 109: Could you please explain why we need this change? Thanks.

fieldservice/views/fsm_person.xml
```diff
@@ -101,6 +101,7 @@
                             <field
                                 name="partner_id"
                                 groups="base.group_no_one"
+                                required="0"
```
cuongnmtm, line 104: and here.

If it is related to the base.group_no_one, maybe we can try to understand why we need it and consider removing it.

hoattk-komit: - The behavior is like the field `partner_id` on the user form view, so I think don't remove it.
- I added `required="0"` because the field is read-only, so we can avoid being blocked on UI.

cuongnmtm: Ah, it is a delegation inheritance. Thanks.

cuongnmtm (approved):
Code review LGTM

cuongnmtm:
@hoattk-komit hoa could you please merge/squash commits? Thanks.

fieldservice/models/fsm_location.py
```diff
@@ -84,17 +85,12 @@ class FSMLocation(models.Model):
         recursive=True,
         store=True,
     )
-    complete_direction = fields.Char(
+    complete_direction = fields.Html(
```
hieulucky111, line 88: Why do we need to change the field type here?

fieldservice/models/fsm_order.py
```diff
@@ -349,10 +354,14 @@ def can_unlink(self):
         """:return True if the order can be deleted, False otherwise"""
         return self.stage_id == self._default_stage_id()
 
+    # pylint: disable=no-raise-unlink
     def unlink(self):
```
hieulucky111: please switch to use `@api.ondelete`.

hieulucky111 (changes requested):

fieldservice/models/fsm_location.py
hoattk-komit: Because in the compute method, this field is computed from the field `direction` (this is a HTML) field.

fieldservice/models/fsm_order.py
hoattk-komit: done

fieldservice/models/fsm_location.py
```diff
@@ -115,7 +111,9 @@ def _compute_complete_name(self):
     def _compute_complete_direction(self):
         for rec in self:
             parent_direction = rec.parent_id.complete_direction
-            complete_direction = (parent_direction or "") + (rec.direction or "")
+            complete_direction = (parent_direction or "<p></p>") + (
```
brian10048: This is not rendering as HTML, but it is showing as plain text:

Image

You can use markupsafe to ensure tags are rendered correctly

brian10048 (changes requested):

fieldservice/models/fsm_location.py
hoattk-komit: Done

hieulucky111 (approved):

brian10048 (approved):

PR #1514, opened 2026-02-15
[17.0] [IMP] fieldservice_account_analytic: use python contrains instead of validation at creation for analytic lines

rrebollo:
I found this issue after seeing the following warning:

`WARNING ? odoo.api.create: The model odoo.addons.fieldservice_account_analytic.models.analytic_account is not overriding the create method in batch`

When refactoring to use `@api.model_create_multi`, it became clear that a Python constraint is the more appropriate solution.

@BinhexTeam T18741


rrebollo:
Failing tests are not related to this addon.

PR #1494, opened 2026-01-08
[ADD] fieldservice_order_property: new module

rrebollo:
This module enables property (`fields.PropertiesDefinition`, `fields.Properties`) management in field service orders. Properties are shared across all field service orders within the same field service team.

@BinhexTeam

PR #1538, opened 2026-05-19
[18.0] [MIG] fieldservice_order_property

rrebollo:
Standard migration from v17.0

#1220 

Based on:

- #1494

@BinhexTeam T18741


rrebollo:
Improved #1510

tonygalmiche (commented):
Tested on Odoo 18.0.
Assigned a team to a fsm.order and saved. The "Add Properties"
button described in USAGE.md never appears on the Properties tab when it is empty —
neither for the very first property, nor after deleting the last existing one (tested
by injecting a property definition directly via SQL, then deleting it through the UI:
the "+" trigger did not come back). 

No other view (e.g. on the Team form) allows
managing/bootstrapping properties either. 
Blocking issue.

rrebollo:
@tonygalmiche did you test it on runboat? Can U upload some screenshoot?

tonygalmiche:
Not a local setup issue — found the root cause.

In the core widget's template
(addons/web/static/src/views/fields/properties/properties_field.xml, line 5),
the whole Properties widget is hidden (`d-none`) when the properties list is
empty AND `showAddButton` is not set:

    t-att-class="{'d-none': propertiesList.length === 0 and !state.showAddButton}"

The view in this PR (views/fsm_order_views.xml) declares the field without
`showAddButton="1"`:

    <field name="fsm_order_properties" columns="2" />

So there's no way to create the first property from an empty state. I confirmed
this by injecting a property definition directly via SQL: the widget then
renders fine, and deleting the last property makes it disappear again — same
issue.

Fix confirmed on my side: adding `showAddButton="1"` makes the "Add Properties"
button appear correctly even with an empty list:

    <field name="fsm_order_properties" columns="2" showAddButton="1" />

Also applies to the kanban view (line 23), which doesn't set it either.

rrebollo:
thank U @tonygalmiche I will fix it ASAP. Then I'll request your review.

rrebollo:
@tonygalmiche solved. Thank you again.

tonygalmiche (approved):
Tested on a local Odoo 18.0 instance:
- The "Add Properties" button now correctly appears on an empty properties list (form and kanban), fixed via showAddButton="1".
- Property definitions are correctly shared across all orders of the same team, while values remain independent per order.
- Kanban view correctly displays the team field and properties (label hidden for tag-type properties, which is standard Odoo core behavior, not specific to this module).
- "Group by Properties" and advanced search filter ("Add Custom Filter") work correctly, listing each property grouped by team.
- Deleting a property removes it for the whole team, as expected.

Works as expected.

Thanks @rrebollo for the quick fix!

PR #1545, opened 2026-06-05
Translated using Weblate (Chinese (China))

bigfoot88:
This PR adds Chinese (China) translation for fieldservice module via weblate-style commit.


pedrobaeza:
Please use https://www.odoo-community.org/resources/translate.

PR #1552, opened 2026-06-22
[19.0][MIG] fieldservice_route: Migration to 19.0

max3903:
Migration of `fieldservice_route` to Odoo 19.0 following OCA maintainer-tools guidelines.

## Changes applied
- Bumped version to `19.0.1.0.0`
- Updated translations for 19.0 (`es`, `es_AR`, `es_CL`, `it`, `pt_BR`, `pt_PT`, `tr`)
- Updated tests to use `FSMCommon` without demo data dependencies
- Updated `self.env._()` translation calls for Odoo 19.0
- Gray Matter Logic credits and readme improvements
- Local pre-commit and module tests passed before push

## Checklist
- [x] Tests pass
- [x] pre-commit passes

Made with [Cursor](https://cursor.com)


brian10048:
@max3903 are you able to include the previous commit history from earlier versions?

max3903:
@Sadiq-OSI @AdriAlpizar

fieldservice_route/security/ir.model.access.csv
```diff
@@ -0,0 +1,9 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_fsm_route_user,fsm.route.user,model_fsm_route,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_dispatcher,fsm.route.dispatcher,model_fsm_route,fieldservice.group_fsm_dispatcher,1,1,1,1
+access_fsm_route_day_user,fsm.route.day.user,model_fsm_route_day,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_day_dispatcher,fsm.route.day.dispatcher,model_fsm_route_day,fieldservice.group_fsm_manager,1,1,0,0
+access_fsm_route_dayroute_user,fsm.route.dayroute.user,model_fsm_route_dayroute,fieldservice.group_fsm_user,1,1,0,0
+access_fsm_route_dayroute_dispatcher,fsm.route.dayroute.dispatcher,model_fsm_route_dayroute,fieldservice.group_fsm_dispatcher,1,1,1,1
+access_fsm_route_dayroute_portal,access.fsm.route.dayroute.portal,model_fsm_route_dayroute,base.group_portal,1,1,1,0
```
Sadiq-OSI: Portal users currently have permission to read, create, and edit all day routes, with no record rule limiting which routes they can access. Could you please restrict this access or add a rule so portal users can only access the appropriate records?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,100 @@
...
+                date = vals.get("scheduled_date_start").date()
+        return {
+            "person_id": vals.get("person_id")
+            or self.person_id.id
+            or self.fsm_route_id.fsm_person_id.id,
+            "date": date or self.scheduled_date_start.date(),
+            "route_id": vals.get("fsm_route_id") or self.fsm_route_id.id,
+        }
+
+    def _get_dayroute_domain(self, values):
```
Sadiq-OSI, line 205: Could you include route_id in this search domain? If the same worker has two routes on the same date, an order for the second route could incorrectly be added to the first route’s day route.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,100 @@
...
+        values = self._get_dayroute_values(vals)
+        domain = self._get_dayroute_domain(values)
+        dayroute = dayroute_obj.search(domain, limit=1)
+        if dayroute:
+            vals.update({"dayroute_id": dayroute.id})
+        else:
+            if self._can_create_dayroute(values):
+                dayroute = dayroute_obj.create(self.prepare_dayroute_values(values))
+                vals.update({"dayroute_id": dayroute.id})
+        # If this was the last order of the dayroute,
```
Sadiq-OSI: This empty-route check happens before the order is moved, so the old day route still contains the order and is never deleted. Could you perform the cleanup after the write, when the old day route is actually empty?

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -0,0 +1,210 @@
...
+    @api.depends("route_id", "route_id.fsm_person_id")
+    def _compute_person_id(self):
+        for rec in self:
+            if not rec.route_id.fsm_person_id:
+                rec.person_id = None
+                continue
+
+            rec.person_id = rec.route_id.fsm_person_id
+
+    @api.depends(
```
Sadiq-OSI, line 155: The planned start uses the worker’s calendar hours and timezone, but changes to those calendar settings do not trigger a recomputation. Could you add the required dependencies so existing day routes are updated when the working schedule changes?

Sadiq-OSI (changes requested):
I found a few issues that could affect security and route assignment consistency. Please review the four inline comments and address them before merging. Thank you.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,107 @@
...
+        return super().create(vals_list)
+
+    def write(self, vals):
+        dayroutes_to_check = self.env["fsm.route.dayroute"]
+        for rec in self:
+            if (vals.get("person_id", False) or rec.person_id) and (
+                vals.get("scheduled_date_start", False) or rec.scheduled_date_start
+            ):
+                dayroutes_to_check |= rec.dayroute_id
+                vals = rec._manage_fsm_route(vals)
```
Sadiq-OSI: When multiple scheduled orders are updated together, vals is reused and changed for every order. The final day route found is then written to all the orders. For example, a simple bulk description update moved an order from Route A onto Route B’s day route.
Please process each order separately, or group orders by the correct day route before calling super().write().

Sadiq-OSI: Clearing scheduled_date_start does not remove the order from its day route. The method uses the order’s previous scheduled date before the write, so the old dayroute_id remains attached even though the order is now unscheduled.
Please clear dayroute_id when scheduled_date_start is explicitly set to False, and delete the old day route if it becomes empty.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,107 @@
...
+        res = super().write(vals)
```
Sadiq-OSI: Route management runs before the new location_id is written, so it still reads the route from the old location. After changing from a Route A location to a Route B location, the order correctly shows Route B but remains attached to Route A’s day route.
Please determine the route from the incoming location_id, or perform the day-route reassignment after the location has been updated.

Sadiq-OSI (changes requested):
I found three routing issues in fsm.order.write() that can leave orders assigned to an incorrect or outdated day route. Please address the inline comments before merging. Apart from that, the module installation and existing test suite otherwise pass.

fieldservice_route/models/fsm_order.py
Sadiq-OSI: Fixed in e8009d8. Each order now receives its own copy of vals, and non-routing bulk updates bypass route reassignment. The added regression test confirms that orders retain their correct day routes. Resolving this thread.

Sadiq-OSI: Fixed in e8009d8. Clearing scheduled_date_start now clears dayroute_id, and the previous day route is deleted when it becomes empty. The new regression test covers this behavior.

fieldservice_route/models/fsm_order.py
Sadiq-OSI: Fixed in e8009d8. Route management now derives the route from the incoming location_id, assigns the order to the correct day route, and removes the previous day route when it becomes empty. The new regression test covers changing from one routed location to another.

fieldservice_route/models/fsm_order.py
Sadiq-OSI: Fixed in e8009d8 for automatic route reassignment. The previous day route is now retained for checking, the order is moved first, and the old day route is then deleted if empty. A separate edge case remains for direct dayroute_id changes, which I will report separately.

fieldservice_route/security/ir.model.access.csv
Sadiq-OSI: portal record rule added. One security issue remains: Odoo checks write access against the current record before applying the new values. A portal user can therefore change route_id or person_id on an owned day route and move it to another worker. Please make portal access read-only, or enforce the permitted fields and ownership server-side, with tests covering cross-worker writes.

fieldservice_route/models/fsm_order.py
Sadiq-OSI: The route is now included when route_id has a value. One case remains: because routes are optional, when route_id is False the condition omits it from the domain. A route-less order can then reuse an unrelated routed day route for the same worker and date. Please include ("route_id", "=", values.get("route_id")), including False, and add a regression test.

fieldservice_route/models/fsm_route_dayroute.py
Sadiq-OSI: The added dependencies cover the timezone and basic attendance hours, but Odoo 19 work intervals also depend on two_weeks_calendar and the attendance fields week_type, day_period, and display_type. Changing those values can leave the stored planned start stale. Please add the corresponding dependencies for each calendar source and include a two-week-calendar regression test.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,149 @@
...
+
+from odoo import api, fields, models
+from odoo.tools import DEFAULT_SERVER_DATETIME_FORMAT
+
+
+class FSMOrder(models.Model):
+    _inherit = "fsm.order"
+
+    _ROUTE_WRITE_FIELDS = frozenset(
+        {"person_id", "scheduled_date_start", "location_id"}
```
Sadiq-OSI: scheduled_date_end also requires route handling. Base fsm.order.write() can derive a new scheduled_date_start when only the end time changes. Because this override bypasses route management for that write, crossing a date boundary leaves the order assigned to its old day route. Please calculate the effective start before the day-route lookup, or reassign after super().write(), and add a regression test.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,149 @@
...
+        ):
+            dayroutes_to_check |= self.dayroute_id
+            write_vals = self._manage_fsm_route(write_vals)
+
+        res = super().write(write_vals)
+        self._unlink_empty_dayroutes(dayroutes_to_check)
+        return res
+
+    def write(self, vals):
+        if not self._ROUTE_WRITE_FIELDS.intersection(vals):
```
Sadiq-OSI: Direct dayroute_id changes bypass this cleanup logic. Moving or removing an order through a day route’s order_ids writes only dayroute_id, so the previous day route can remain empty. Please capture the previous day route and delete it after the write when it becomes empty, with tests for direct removal and movement.

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @api.model
+    def _planned_start_from_date(self, route_date, person=None, route=None):
+        """Return planned start as UTC naive datetime for the given route date."""
+        route_date = fields.Date.to_date(route_date)
+        if route and not person:
+            person = route.fsm_person_id
+        calendar = (
+            person and person.calendar_id
+        ) or self.env.company.resource_calendar_id
```
Sadiq-OSI: The compute dependencies monitor team_id.company_id.resource_calendar_id, but this fallback reads env.company.resource_calendar_id. In a multi-company environment, a day route using a team from another allowed company can therefore receive a planned start from the wrong calendar. Please calculate the fallback from the day route team’s company and keep the computation and dependencies aligned.

Sadiq-OSI (changes requested):
The update fixed a few things. The bulk multi-record write, explicit unscheduling, location-based reassignment, and cleanup after automatic route moves are now fixed.
Some blocking security and routing cases remain. Portal writes can still escape the ownership rule, route-less orders can reuse an unrelated routed day route, and some date or direct day-route changes bypass reassignment or cleanup. Planned-start calendar handling also remains incomplete.
Please address the remaining conversations and new inline comments, and add focused regression tests.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("fsm_route_id") and vals.get("location_id"):
+                location = self.env["fsm.location"].browse(vals["location_id"])
+                vals.update({"fsm_route_id": location.fsm_route_id.id})
+
+            if vals.get("person_id") and vals.get("scheduled_date_start"):
+                vals = self._manage_fsm_route(vals)
```
Sadiq-OSI, line 288: An explicitly provided dayroute_id is overwritten here by _manage_fsm_route(). This is confirmed by the current CI failure: test_dayroute_order_count provides a day route, but that route keeps order_count = 0. Please preserve an explicit day-route selection and make both Odoo and OCB test jobs pass.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,174 @@
...
+        return super().create(vals_list)
+
+    def _prepare_route_write_vals(self, vals):
+        write_vals = dict(vals)
+        if {
+            "scheduled_date_end",
+            "scheduled_duration",
+            "scheduled_date_start",
+        } & write_vals.keys():
+            self._calc_scheduled_dates(write_vals)
```
Sadiq-OSI, line 298: This calculates the scheduled dates here, but the inherited fsm.order.write() calculates them again. For example, setting a two-hour order starting Tuesday at 00:30 to duration 0 can select Tuesday’s day route during the first calculation, then store Monday 22:30 during the second calculation. Please calculate the dates only once and add a regression test for zero duration across a date boundary.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,174 @@
...
+        if self._is_unscheduling(write_vals):
+            dayroutes_to_check |= self.dayroute_id
+            write_vals["dayroute_id"] = False
+            res = super().write(write_vals)
+            self._unlink_empty_dayroutes(dayroutes_to_check)
+            return res
+
+        if "dayroute_id" in write_vals:
+            dayroutes_to_check |= self.dayroute_id
+            res = super().write(write_vals)
```
Sadiq-OSI, line 385: Direct assignment accepts any day route without checking its date, route, or worker. The added test_write_direct_dayroute_id_clears_old_dayroute demonstrates this by assigning an order to a day route seven days later while leaving the order’s scheduled date unchanged. Please validate compatibility or synchronize the order fields when moving it directly.

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -0,0 +1,234 @@
...
+                )
+                person = (
+                    self.env["fsm.person"].browse(vals["person_id"])
+                    if vals.get("person_id")
+                    else self.env["fsm.person"]
+                )
+                team = (
+                    self.env["fsm.team"].browse(vals["team_id"])
+                    if vals.get("team_id")
+                    else self.env["fsm.team"]
```
Sadiq-OSI: When team_id is omitted, this calculates the planned start with an empty team, so _planned_start_from_date() falls back to env.company before super().create() applies the actual default team. Because date_start_planned is explicitly supplied, that value can remain based on the wrong company calendar. Please calculate it after defaults are applied, or resolve the actual default team first, and test creation without an explicit team_id.

Sadiq-OSI (changes requested):
The latest update fixes the module import error and allows the tests to run. However, both Odoo and OCB checks remain failing, with a confirmed day-route assignment failure. I also found issues with double scheduled-date calculation, inconsistent direct day-route assignments, and planned-start calculation when the default team is used. Please address the inline comments and ensure both test jobs pass.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,178 @@
...
+        dayroutes_to_check = self.env["fsm.route.dayroute"]
+
+        if self._is_unscheduling(write_vals):
+            dayroutes_to_check |= self.dayroute_id
+            write_vals["dayroute_id"] = False
+            res = super().write(write_vals)
+            self._unlink_empty_dayroutes(dayroutes_to_check)
+            return res
+
+        if "dayroute_id" in write_vals:
```
Sadiq-OSI, line 379: This direct dayroute_id branch accepts the new day route without checking that its date and route match the order. A user could attach a Monday order to a Tuesday day route, leaving the order schedule and day route inconsistent. Can we validate the target day route or synchronize the order schedule when dayroute_id is written directly?

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -0,0 +1,234 @@
...
+    max_order = fields.Integer(
+        related="route_id.max_order",
+        string="Maximum Capacity",
+        store=True,
+        help="Maximum numbers of orders that can be added to this day route.",
+    )
+
+    def _default_team_id(self):
+        teams = self.env["fsm.team"].search(
+            [("company_id", "in", (self.env.user.company_id.id, False))],
```
Sadiq-OSI: This uses env.user.company_id instead of env.company. In multi-company use, a user working in another active company could get a day route defaulted to the wrong company’s team/calendar. Can this follow the base FSM default pattern and search using the active company?

fieldservice_route/security/ir.model.access.csv
```diff
@@ -0,0 +1,9 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_fsm_route_user,fsm.route.user,model_fsm_route,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_dispatcher,fsm.route.dispatcher,model_fsm_route,fieldservice.group_fsm_dispatcher,1,1,1,1
+access_fsm_route_day_user,fsm.route.day.user,model_fsm_route_day,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_day_dispatcher,fsm.route.day.dispatcher,model_fsm_route_day,fieldservice.group_fsm_manager,1,1,0,0
+access_fsm_route_dayroute_user,fsm.route.dayroute.user,model_fsm_route_dayroute,fieldservice.group_fsm_user,1,1,0,0
+access_fsm_route_dayroute_dispatcher,fsm.route.dayroute.dispatcher,model_fsm_route_dayroute,fieldservice.group_fsm_dispatcher,1,1,1,1
+access_fsm_route_dayroute_portal,access.fsm.route.dayroute.portal,model_fsm_route_dayroute,base.group_portal,1,0,0,0
+access_ir_sequence_portal,access.ir.sequence.portal,base.model_ir_sequence,base.group_portal,1,0,0,0
```
Sadiq-OSI: This grants portal users read access to ir.sequence, which exposes internal sequence configuration beyond what is needed to view their own day routes. Can we remove this access unless there is a specific portal flow that requires it?

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -0,0 +1,234 @@
...
+                match_end=False,
+                search_range=(day_start, day_end),
+                compute_leaves=False,
+            )
+            if work_time:
+                return work_time.astimezone(utc).replace(tzinfo=None)
+        fallback = tzinfo.localize(datetime.combine(route_date, time(8, 0)))
+        return fallback.astimezone(utc).replace(tzinfo=None)
+
+    @api.depends("route_id", "order_ids")
```
Sadiq-OSI: order_remaining is computed from rec.max_order, but the dependency list does not include max_order or route_id.max_order. If the route capacity changes, existing day routes can keep stale remaining capacity and the assignment domain may reuse or skip routes incorrectly. Can we add the missing dependency?

Sadiq-OSI (changes requested):
The module is mostly in good shape and the main test/pre-commit checks are passing, but I found a few issues that should be addressed.

The main concern is that direct dayroute_id writes can attach an order to a day route with a different date or route. I also noticed a multi-company defaulting issue where day routes use env.user.company_id instead of the active company, portal users are granted more ir.sequence read access than needed, and remaining capacity may become stale when a route’s max_order changes.

There are also failing Codecov checks. They seem coverage-related rather than functional test failures, but adding focused tests for these cases would likely help with both the review issues and the coverage gates and will also pass those codecov checks that are currently failing.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,234 @@
...
+    person_id = fields.Many2one(
+        comodel_name="fsm.person",
+        string="Assigned To",
+        index=True,
+        compute="_compute_person_id",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("fsm_route_id")
```
Sadiq-OSI: Changing fsm.route.fsm_person_id does not invalidate this stored compute because the dependency only watches fsm_route_id. The day route already depends on route_id.fsm_person_id, so changing a route worker leaves the order assigned to the old person while its day route moves to the new person.

Could you include fsm_route_id.fsm_person_id in the dependency and add a regression test for changing a route's worker?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def _get_dayroute_values(self, vals):
+        date = False
+        if vals.get("scheduled_date_start"):
+            if isinstance(vals.get("scheduled_date_start"), str):
+                date = datetime.strptime(
+                    vals.get("scheduled_date_start"), DEFAULT_SERVER_DATETIME_FORMAT
+                ).date()
+            elif isinstance(vals.get("scheduled_date_start"), datetime):
+                date = vals.get("scheduled_date_start").date()
```
Sadiq-OSI: scheduled_date_start is stored as a UTC-naive datetime, so calling .date() here selects the UTC day rather than the worker's local route day. I reproduced a Toronto order scheduled Tuesday at 23:30 being treated as Wednesday and rejected by a Tuesday-only route.

Could you convert the scheduled start to the worker or route timezone before extracting the date and add a boundary regression test?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,234 @@
...
+    def create(self, vals_list):
+        for vals in vals_list:
+            if not vals.get("fsm_route_id") and vals.get("location_id"):
+                location = self.env["fsm.location"].browse(vals["location_id"])
+                vals.update({"fsm_route_id": location.fsm_route_id.id})
+
+            if (
+                vals.get("person_id")
+                and vals.get("scheduled_date_start")
+                and "dayroute_id" not in vals
```
Sadiq-OSI: Skipping _manage_fsm_route() preserves an explicitly provided dayroute_id, but the create path never calls _prepare_vals_from_dayroute(). Creating an order from a day route can therefore leave scheduled_date_start empty, and API creates can accept a day route whose route, date, or person does not match the order.

Could you validate and synchronize explicit day routes during create and add regression tests?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,234 @@
...
+                    "The day route %(dayroute)s belongs to route %(route)s, "
+                    "which does not match the order route %(order_route)s.",
+                    dayroute=dayroute.display_name,
+                    route=dayroute.route_id.display_name,
+                    order_route=order_route.display_name,
+                )
+            )
+        if dayroute.person_id and "person_id" not in write_vals:
+            write_vals["person_id"] = dayroute.person_id.id
+        if dayroute.date and "scheduled_date_start" not in write_vals:
```
Sadiq-OSI: This leaves a remaining gap in the earlier direct-assignment review. _prepare_vals_from_dayroute() copies the day-route worker and date only when those fields are absent. A write that supplies dayroute_id together with another person_id or scheduled_date_start succeeds and leaves the order inconsistent with the selected day route.

Could you either make the day route authoritative or reject conflicting values, with a combined-write regression test?

Sadiq-OSI (changes requested):
The earlier portal security, route-domain, cleanup, calendar, multi-company, capacity, and CI test issues are largely addressed. I found four remaining consistency cases in fsm.order:

- route worker changes do not recompute the stored order assignee;
- route-day selection uses the UTC date instead of the worker's local date;
- explicit day routes are not validated or synchronized during create;
- combined direct writes can preserve person or date values that conflict with the selected day route.

Please address the inline comments and add focused regression tests before merging. The Odoo, OCB, and pre-commit checks pass, while codecov/project remains failing.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,273 @@
...
+        if current_start:
+            local = utc.localize(fields.Datetime.to_datetime(current_start)).astimezone(
+                tzinfo
+            )
+            local = local.replace(
+                year=dayroute.date.year,
+                month=dayroute.date.month,
+                day=dayroute.date.day,
+            )
+            return local.astimezone(utc).replace(tzinfo=None)
```
Sadiq-OSI: local.replace() keeps the timezone offset from the original date. For example, moving a Toronto appointment at 10:00 AM from July to January results in 9:00 AM because the July UTC offset is retained. Can you re-localize the local time using the target date so the displayed appointment time remains 10:00 AM? Please add a regression test that moves an order across a daylight-saving boundary.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,273 @@
...
+            return self.env["fsm.location"].browse(location_id).fsm_route_id
+        return self.fsm_route_id
+
+    def _prepare_vals_from_dayroute(self, dayroute, write_vals):
+        """Make an explicit day route authoritative for person and schedule date."""
+        write_vals = dict(write_vals)
+        if not dayroute:
+            return write_vals
+        order_route = self._order_route_for_dayroute_write(write_vals)
+        if dayroute.route_id and order_route and dayroute.route_id != order_route:
```
Sadiq-OSI: This validation only rejects a mismatch when both the order and day route have a route. A Route A order can still be assigned to a day route with no route, leaving the records inconsistent. Could we compare the route values even when one side is empty and reject one-sided mismatches? Please add regression tests for both routed-order -> route-less-day-route and route-less-order -> routed-day-route.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,273 @@
...
+    person_id = fields.Many2one(
+        comodel_name="fsm.person",
+        string="Assigned To",
+        index=True,
+        compute="_compute_person_id",
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("fsm_route_id", "fsm_route_id.fsm_person_id")
```
Sadiq-OSI, line 41: Changing the route worker now recomputes the order’s person_id, but it does not re-evaluate the existing day-route date using the new worker’s timezone. For example, changing from a Toronto worker to a Los Angeles worker can make the appointment fall on the previous local day while it remains attached to the original day route. Could we synchronize or reassign the day route when the route worker changes and add a timezone-boundary regression test?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,273 @@
...
+            res = super().write(write_vals)
+            self._unlink_empty_dayroutes(dayroutes_to_check)
+            return res
+
+        if "dayroute_id" in write_vals:
+            dayroutes_to_check |= self.dayroute_id
+            new_dayroute = self.env["fsm.route.dayroute"].browse(
+                write_vals.get("dayroute_id") or []
+            )
+            write_vals = self._prepare_vals_from_dayroute(new_dayroute, write_vals)
```
Sadiq-OSI, line 384: When dayroute_id is written together with scheduled_duration or scheduled_date_end, the end time is calculated from the old start date before this code moves the start to the new day route. I reproduced an order being stored with a January 22 start, January 15 end, and a duration of -166 hours. Could you calculate the end and duration after establishing the new day-route start, and add a regression test for this combined write?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,273 @@
...
+                    order_route=order_route.display_name,
+                )
+            )
+        if dayroute.person_id:
+            write_vals["person_id"] = dayroute.person_id.id
+        if dayroute.date:
+            current_start = (
+                write_vals.get("scheduled_date_start") or self.scheduled_date_start
+            )
+            write_vals["scheduled_date_start"] = self._scheduled_start_on_dayroute_date(
```
Sadiq-OSI, line 352: Changing dayroute_id updates scheduled_date_start, but it does not correctly update scheduled_date_end. I reproduced a two-hour 09:00–11:00 order becoming 09:00–09:00 while still storing a two-hour duration. Could you keep the start, end, and duration synchronized and add a regression test for moving an existing order to another day route?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,273 @@
...
+            "date": date,
+            "route_id": route_id,
+        }
+
+    def _get_dayroute_domain(self, values):
+        return [
+            ("person_id", "=", values["person_id"]),
+            ("date", "=", values["date"]),
+            ("route_id", "=", values.get("route_id") or False),
+            ("order_remaining", ">", 0),
```
Sadiq-OSI, line 211: This capacity condition excludes the order’s current day route when that route is full. As a result, simply changing the order’s duration creates a replacement day route and deletes the original one. Could you preserve the current matching day route because the order already occupies one of its slots, and add a test using max_order = 1?

Sadiq-OSI (changes requested):

fieldservice_route/security/ir.model.access.csv
```diff
@@ -0,0 +1,8 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_fsm_route_user,fsm.route.user,model_fsm_route,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_dispatcher,fsm.route.dispatcher,model_fsm_route,fieldservice.group_fsm_dispatcher,1,1,1,1
+access_fsm_route_day_user,fsm.route.day.user,model_fsm_route_day,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_day_dispatcher,fsm.route.day.dispatcher,model_fsm_route_day,fieldservice.group_fsm_manager,1,1,0,0
+access_fsm_route_dayroute_user,fsm.route.dayroute.user,model_fsm_route_dayroute,fieldservice.group_fsm_user,1,1,0,0
```
Sadiq-OSI: Regular Field Service users can reschedule orders, but the automatic day-route creation fails with an AccessError because they don’t have create access. Please update the permissions and add a test for this workflow.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,326 @@
...
+        if dayroute.date:
+            current_start = (
+                write_vals.get("scheduled_date_start") or self.scheduled_date_start
+            )
+            write_vals["scheduled_date_start"] = self._scheduled_start_on_dayroute_date(
+                current_start,
+                dayroute,
+                person=dayroute.person_id,
+            )
+            write_vals = self._sync_scheduled_end_from_start(write_vals)
```
Sadiq-OSI, line 357: Creating an order with a day route, start time, and end time resets the end time to the start and sets the duration to zero. Please calculate the duration from the supplied start and end before synchronizing the dates, and add a regression test.

fieldservice_route/models/fsm_route.py
```diff
@@ -0,0 +1,42 @@
...
+            return day in self.day_ids
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "fsm_person_id" in vals:
+            orders = self.env["fsm.order"].search(
+                [
+                    ("fsm_route_id", "in", self.ids),
+                    ("scheduled_date_start", "!=", False),
+                    ("dayroute_id", "!=", False),
```
Sadiq-OSI: When a worker is added to a previously unstaffed route, scheduled orders without a day route are skipped by this filter. The worker is assigned, but the orders remain outside any day route. Please include these orders in the reassignment.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,326 @@
...
+    def _get_route_id_from_vals(self, vals):
+        if vals.get("fsm_route_id"):
+            return vals["fsm_route_id"]
+        location_id = vals.get("location_id") or self.location_id.id
+        if location_id:
+            return self.env["fsm.location"].browse(location_id).fsm_route_id.id
+        return self.fsm_route_id.id
+
+    def _get_person_id_for_dayroute(self, vals, route_id):
+        if vals.get("person_id"):
```
Sadiq-OSI: Setting person_id to false falls back to the previous worker, so the order remains attached to that worker’s day route after being unassigned. Please handle an explicit false value separately and clear the obsolete day route.

Sadiq-OSI (changes requested):

fieldservice_route/security/ir.model.access.csv
```diff
@@ -0,0 +1,8 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_fsm_route_user,fsm.route.user,model_fsm_route,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_dispatcher,fsm.route.dispatcher,model_fsm_route,fieldservice.group_fsm_dispatcher,1,1,1,1
+access_fsm_route_day_user,fsm.route.day.user,model_fsm_route_day,fieldservice.group_fsm_user,1,0,0,0
+access_fsm_route_day_dispatcher,fsm.route.day.dispatcher,model_fsm_route_day,fieldservice.group_fsm_manager,1,1,0,0
+access_fsm_route_dayroute_user,fsm.route.dayroute.user,model_fsm_route_dayroute,fieldservice.group_fsm_user,1,1,1,0
```
Sadiq-OSI, line 6: The create access fixes the original error, but regular Field Service users still get an AccessError when rescheduling an order that already has a day route. The cleanup tries to delete the old empty day route, but these users don’t have unlink access. Could yopu handle that cleanup with the appropriate permissions and add a test for rescheduling an already routed order?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,353 @@
...
+        elif self._can_create_dayroute(values):
+            dayroute = dayroute_obj.create(self.prepare_dayroute_values(values))
+            vals.update({"dayroute_id": dayroute.id})
+        return vals
+
+    def _reassign_dayroute_for_person_timezone(self):
+        """Re-evaluate dayroute after the worker/timezone changes."""
+        self.ensure_one()
+        if not self.scheduled_date_start or not self.person_id:
+            return
```
Sadiq-OSI: Clearing the worker from a route still leaves its scheduled orders attached to the existing day route. This returns early when the order no longer has a worker, so the old day-route assignment is never cleared. Could you clean it up here and add a test for removing a route’s worker?

fieldservice_route/models/fsm_location.py
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (C) 2019 Serpent consulting Services
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class FSMLocation(models.Model):
+    _inherit = "fsm.location"
+
+    fsm_route_id = fields.Many2one(comodel_name="fsm.route", string="Route")
```
Sadiq-OSI, line 11: Changing a location’s route updates the scheduled order’s route and worker, but its day route still belongs to the previous route. Could you reassign existing scheduled orders when the location route changes and add a regression test for this case?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,353 @@
...
+            and (dayroute.route_id.id or False) == (values.get("route_id") or False)
+        )
+
+    def _can_create_dayroute(self, values):
+        return values["person_id"] and values["date"]
+
+    def _unlink_empty_dayroutes(self, dayroutes):
+        dayroutes.filtered(
+            lambda dayroute: dayroute and not dayroute.order_ids
+        ).unlink()
```
Sadiq-OSI: When a regular Field Service user reschedules an order that already has a day route, this cleanup tries to delete the old day route. Those users do not have delete access, so the reschedule fails with an AccessError. Could you handle this automatic cleanup with the required permissions and add a test for this workflow?

Sadiq-OSI (changes requested):

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,358 @@
...
+        return self.fsm_route_id.id
+
+    def _get_person_id_for_dayroute(self, vals, route_id):
+        if "person_id" in vals:
+            return vals.get("person_id") or False
+        if route_id:
+            route = self.env["fsm.route"].browse(route_id)
+            if route.fsm_person_id:
+                return route.fsm_person_id.id
+        return self.person_id.id or self.fsm_route_id.fsm_person_id.id
```
Sadiq-OSI: When an order is moved to a location with an unstaffed route, it keeps the previous worker and creates a stale day route. Please clear the day route in this case and add a regression test.

fieldservice_route/models/fsm_route_dayroute.py
```diff
@@ -0,0 +1,234 @@
...
+        calendar = (person and person.calendar_id) or company.resource_calendar_id
+        tz_name = (
+            (person and person.partner_id.tz)
+            or (calendar and calendar.tz)
+            or self.env.user.tz
+            or "UTC"
+        )
+        tzinfo = timezone(tz_name)
+        day_start = tzinfo.localize(datetime.combine(route_date, time.min))
+        day_end = day_start + relativedelta(days=1)
```
Sadiq-OSI: Adding one day here keeps the old UTC offset across DST changes, which can exclude valid working hours. Please build the end time from the next local midnight and add tests for both DST transitions.

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,358 @@
...
+            return vals["fsm_route_id"]
+        location_id = vals.get("location_id") or self.location_id.id
+        if location_id:
+            return self.env["fsm.location"].browse(location_id).fsm_route_id.id
+        return self.fsm_route_id.id
+
+    def _get_person_id_for_dayroute(self, vals, route_id):
+        if "person_id" in vals:
+            return vals.get("person_id") or False
+        if route_id:
```
Sadiq-OSI, line 95: When an order is manually assigned to someone other than the route’s default worker, rescheduling it moves it to the default worker’s day route while keeping the manually assigned worker on the order. Could you preserve the order’s current worker unless the route or location changes, and add a regression test for this case?

Sadiq-OSI (changes requested):
Updating my review status to Request changes for the issues noted earlier

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,380 @@
...
+        store=True,
+        readonly=False,
+    )
+
+    @api.depends("fsm_route_id", "fsm_route_id.fsm_person_id")
+    def _compute_person_id(self):
+        for item in self.filtered("fsm_route_id"):
+            item.person_id = item.fsm_route_id.fsm_person_id
+
+    def prepare_dayroute_values(self, values):
```
Sadiq-OSI, line 46: Could you include team_id when matching and creating day routes? Currently, scheduling a Company B order while Company A is active can create or reuse Company A’s day route, which may use the wrong company calendar and planned start time. Please also add a multi-company regression test for this case.

Sadiq-OSI (changes requested):

fieldservice_route/models/fsm_order.py
max3903: @Sadiq-OSI Fixed!

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,400 @@
...
+            route = self.env["fsm.route"].browse(route_id)
+            if route.fsm_person_id:
+                return route.fsm_person_id.id
+        return self.fsm_route_id.fsm_person_id.id
+
+    def _tz_name_for_route_day(self, person=None, route=None):
+        person = person or self.env["fsm.person"]
+        route = route or self.env["fsm.route"]
+        return (
+            (person.partner_id.tz if person else False)
```
Sadiq-OSI, line 106: Small edge case: if the worker’s partner has no timezone but their calendar does, this falls back to the current user’s timezone and can select the wrong day route around midnight. Could you use the worker calendar timezone before the user fallback and add a test for this case?

fieldservice_route/models/fsm_order.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def _prepare_vals_from_dayroute(self, dayroute, write_vals):
+        """Make an explicit day route authoritative for person and schedule date."""
+        write_vals = dict(write_vals)
+        if not dayroute:
+            return write_vals
+        order_route = self._order_route_for_dayroute_write(write_vals)
+        dayroute_route_id = dayroute.route_id.id if dayroute.route_id else False
+        order_route_id = order_route.id if order_route else False
+        if dayroute_route_id != order_route_id:
```
Sadiq-OSI, line 328: One more edge case: an order and day route can belong to different teams as long as they use the same route. This leaves the order attached to a day route with a different team and calendar. Could you validate or synchronize the team here?

Sadiq-OSI (approved):
The migration installs successfully and the full module test suite passes. I found two edge cases around timezone fallback and cross-team day-route assignment, but neither needs to block this migration and both can be handled in a follow-up PR.

PR #1553, opened 2026-06-22
[16.0][FIX] fieldservice_isp_flow: fix search view inheritance with translations

### LINKED ISSUE #1525: fieldservice_isp_flow: search view inheritance broken when stage names are translated
## Bug description

The search view inheritance in `fieldservice_isp_flow` uses `<search string="Orders">` to locate the parent element, but when the base `fieldservice` module has its stage/view strings translated (e.g. to French: "Commandes"), the match fails with:

```
ValueError: L'élément '<search string="Orders">' ne peut être localisé dans la vue parente
```

This makes `fieldservice_isp_flow` unusable on any database with translations loaded.

## Steps to reproduce

1. Install `fieldservice` and load French translations
2. Install `fieldservice_isp_flow`
3. Navigate to Field Service → Operations → All Orders
4. Error: `ValueError: Element '<search string="Orders">' cannot be located in parent view`

## Expected behavior

The view inheritance should work regardless of translations.

## Proposed fix

In `fieldservice_isp_flow/views/fsm_order.xml`, replace the string-based match with an explicit XPath:

```xml
<!-- Before (broken with translations) -->
<search string="Orders">
    <filter name="order_today" position="before">
        ...
    </filter>
</search>

<!-- After (works with any language) -->
<xpath expr="//filter[@name='order_today']" position="before">
    ...
</xpath>
```

## Environment

- Odoo version: 16.0
- Module version: 16.0.1.0.0
- Language: French (fr_FR)

max3903:
## Summary

- Replace string-based search view inheritance (`<search string="Orders">`) with an XPath targeting `//filter[@name='order_today']`
- Ensures `fieldservice_isp_flow` works when parent view labels are translated (e.g. French)

Fixes https://github.com/OCA/field-service/issues/1525

## Test plan

- [x] Install `fieldservice` and load French translations
- [x] Install `fieldservice_isp_flow`
- [x] Open Field Service → Operations → All Orders
- [x] Confirm the search view loads without error and the "Past Due Orders" filter is present

Made with [Cursor](https://cursor.com)

### DIFF
--- a/fieldservice_isp_flow/views/fsm_order.xml
+++ b/fieldservice_isp_flow/views/fsm_order.xml
@@ -93,15 +93,13 @@
         <field name="model">fsm.order</field>
         <field name="inherit_id" ref="fieldservice.fsm_order_search_view" />
         <field name="arch" type="xml">
-            <search string="Orders">
-                <filter name="order_today" position="before">
-                    <filter
-                        string="Past Due Orders"
-                        domain="[('stage_id.name', 'not in', ['Started', 'Completed', 'Cancelled']), ('scheduled_date_start', '&lt;', current_date)]"
-                        name="past_due"
-                    />
-                </filter>
-            </search>
+            <xpath expr="//filter[@name='order_today']" position="before">
+                <filter
+                    string="Past Due Orders"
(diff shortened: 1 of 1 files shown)


max3903:
@risto42

PR #1570, opened 2026-07-17
[19.0][MIG] fieldservice_size: Migration to 19.0

LINTOANTONY007:


LINTOANTONY007:
@max3903  , Can you review this PR , Thanks in advance!

max3903 (changes requested):
Thanks for the migration. A few issues stood out in review:

fsm.order._compute_size_value — inside the for rec in self loop the domain uses self.location_id / self.size_id instead of rec.location_id / rec.size_id. That will break (or mis-assign) when several orders are computed in one batch.

fsm.size._one_size_per_type — same pattern: self.type_id.id is used on the whole recordset. Please loop over self (or use a domain that covers all records) so multi-record create/write does not raise a singleton error.

_compute_size_id — when type is cleared, size_id is never reset, so the order can keep a stale default size.

Tests — in test_order_onchange_location, assertTrue(order.size_id, self.size_a.id) (and the similar checks) do not assert equality; the second argument is only a failure message. Prefer assertEqual.

Happy to re-review once those are addressed.

LINTOANTONY007:
@max3903 , Thanks for the review.You are absolutely right. I wonder how I missed these points. Anyway, I did those changes in code, Can you re-review ?

max3903 (changes requested):
Thanks for addressing the previous review points — those look good.

Two leftover issues in _compute_size_value:

- [x] When size_id or location_id is missing, the method continues without resetting size_value, so a stored quantity can remain after the type/size is cleared.
- [x] The compute only depends on size_id / location_id, not on the related fsm.location.size quantity, so editing or removing a location size line leaves the order’s size_value stale.

Otherwise LGTM from my side once those are handled.

LINTOANTONY007:
@max3903 , Thanks for the feedbacks,Done the changes, Please review.

max3903 (approved):

PR #1462, opened 2025-11-03
[18.0][MIG] fieldservice_stage_validation: Migration to 18.0

fsmw:
Migration of fieldservice_stage_validation module to Odoo 18.0.

## Summary
- ✅ **Proper migration using git format-patch method** - preserves complete git history from 17.0 branch
- ✅ Updated module version to 18.0.1.0.0  
- ✅ **Fixed translation issue** - use rec.env._ for proper translation context
- ✅ Applied all pre-commit fixes and formatting
- ✅ Follows OCA migration guidelines exactly

## Module Description
This module adds field validation to FSM stages, allowing administrators to configure required fields that must be set before moving to specific stages in the field service workflow.

## Migration Details
- **Method**: Used git format-patch to preserve complete git history (not manual copy)
- **Code Changes**: Updated translation calls to use proper rec.env._ context
- **Quality**: All pre-commit linting and formatting applied
- **Testing**: Ready for Runboat CI testing environment

## Technical Changes
- Fixed: "no translation language detected" warning in validate_utils.py
- Updated: Manifest version format to 18.0.1.0.0
- Preserved: Full commit history from 17.0 branch

Closes part of migration checklist in OCA/field-service#1220

PR #1327, opened 2025-02-21
[16.0][FIX]fieldservice_stage_validation: allow fieldservice manager to change fields

PicchiSeba:
In `14.0` a user in the group `fieldservice.manager` could change the "Field to Validate" in Field Service > Configuration > Stages.

In `16.0` this is not the case anymore, because the group is missing the read permissions on `ir.model` and `ir.model.fields`

In this PR we restore read permissions while setting some rules to limit them to the relevant tables only.

PR #1576, opened 2026-07-22
[MIG] fieldservice_stage_validation: Migration to 18.0

SirPyTech:
Migration from `17.0` (https://github.com/OCA/field-service/tree/48857bdb6a2fc7ea128fbd7bc64f0d4722bb6d57/fieldservice_stage_validation).

@fsmw you wanted to migrate this module in https://github.com/OCA/field-service/pull/1462 but your PR went stale, maybe you want to have a look?

@EdgarRetes you wanted to migrate this module in https://github.com/OCA/field-service/pull/1248 but your PR went stale, maybe you want to have a look?

I added a fix to the domain for fields of a Worker's Stage: without the fix, the fields of model `fsm.worker` are searched but that model does not exist (should be `fsm.person`).
The fix is in a separated commit so it can be easily backported.

I also added a fix for the issue described in https://github.com/OCA/field-service/pull/1327: when a FieldService Manager opens a Stage the following error is shown:
> You are not allowed to access 'Models' (ir.model) records.
> This operation is allowed for the following groups:
> 	- Administration/Access Rights
> 
> Contact your administrator to request access if necessary.

The added fixes are not strictly part of the migration and I can move them to another PR if needed, please let me know.


max3903:
@SirPyTech This module should not even exist. You can achieve this with `fieldservice_stage_server_action`.

SirPyTech:
> @SirPyTech This module should not even exist. You can achieve this with `fieldservice_stage_server_action`.

Thanks for the hint! Indeed the server actions could achieve what this module is doing, and more.

On the other hand: this module is way easier to manage for the end user, which might not be able to create/edit the Python code needed for server action.

HekkiMelody (approved):
Code review, LGTM

max3903 (approved):

PR #1578, opened 2026-07-22
[18.0][MIG] fieldservice_account_analytic

Jimmy-sc94:
Migration of fieldservice_account_analytic to 18.0.

Supersedes #1226 (closed as stale). This is a plain migration of the 17.0 module, as suggested in that PR discussion: the extra features that raised review concerns there (analytic plans, automatic analytic account creation loops, fsm.order.cost, the account_usability dependency) are NOT included.

Review feedback from #1226 addressed on the 17.0 codebase:
- account.move.line analytic assignment now extends _compute_analytic_distribution (no create override), with a constrains keeping the 17.0 validation.
- Record rules moved to security/ with tightened domains and a read-only dispatcher rule.
- ACLs reduced to minimal privilege; obsolete onchange removed.
- create() is @api.model_create_multi with a guard on fsm_order_id.

18.0 adaptations verified against the 18.0 branches: fsm_order_list_view / fsm_location_list_view inheritance ids, _search signature without access_rights_uid, t-out in report templates, demo data re-pointed to analytic.analytic_plan_projects, tests updated (no detailed_type, fields.Command, tracking_disable).

Module installs on a real Odoo 18 database and all its tests pass with demo data.


Jimmy-sc94:
Note for reviewers: this PR contains a second commit touching fieldservice_subcontracting tests only. That module is new in 18.0 and its tests create analytic lines linked to FSM orders whose test location has no analytic account — which this module (faithful to its 17.0 behavior) rejects. The fix seeds an analytic account on the test location, guarded so those tests keep passing when this module is not installed.

PR #1608, opened 2026-08-17
[18.0][FIX] fieldservice_route: one dayroute per worker and day

tortilicious:
## Problem

In production, technicians end up with several `fsm.route.dayroute` records for the same day, some of them empty/orphaned. Three independent defects in `fsm.order`/`fsm.route.dayroute` compound into this:

**1. `write()` re-runs dayroute management on every write, regardless of what changed**

```python
def write(self, vals):
    for rec in self:
        ...
        if (vals.get("person_id", False) or rec.person_id) and (
            vals.get("scheduled_date_start", False) or rec.scheduled_date_start
        ):
            vals = rec._manage_fsm_route(vals)
    return super().write(vals)
```

The condition is true for *any* write on an order that already has a technician and a scheduled date, even if `vals` doesn't touch either. Combined with defect 3 below, an unrelated write on an order whose route is already at capacity kicks it out into a brand new dayroute for no reason — this reproduces exactly the "writes right after create()" pattern from modules like `fieldservice_calendar`.

The same loop also mutates and reassigns the *same* `vals` dict shared across `self`, then calls `super().write(vals)` **once**, outside the loop, with whatever `vals` looks like after the last iteration — so in a multi-record write mixing orders of different technicians, every record ends up written with the *last* record's resolved `dayroute_id`.

There's also a dead branch handling `vals.get("route_id")` / `route.date`: neither `route_id` nor `date` exist on `fsm.order`/`fsm.route` respectively, so this can never execute (or raises `AttributeError` if it ever did).

**2. The empty-dayroute cleanup runs before the order has actually moved**

```python
def _manage_fsm_route(self, vals):
    ...
    if self.dayroute_id and not self.dayroute_id.order_ids:
        self.dayroute_id.unlink()
    return vals
```

This check happens *before* `super().write(vals)` applies the new `dayroute_id`, so `self.dayroute_id.order_ids` still contains the order that is about to leave — the condition is never true when it should be, and the old (now-empty) dayroute is never deleted when an order changes day/technician.

**3. `max_order = 0` (the field's own default) makes the domain unsatisfiable**

```python
def _get_dayroute_domain(self, values):
    return [
        ("person_id", "=", values["person_id"]),
        ("date", "=", values["date"]),
        ("order_remaining", ">", 0),
    ]
```

`order_remaining = max_order - order_count`, and `max_order` defaults to 0 (`fsm.route.max_order`, "Maximum number of orders per day route"). With no explicit capacity configured, `order_remaining` is never positive, so the existing dayroute is never found again and a new one gets created on every write. `check_capacity()` has the same asymmetry the other way: with `max_order = 0`, a *single* order already violates it.

## Fix

- `write()` only touches dayroute management when `vals` actually contains `person_id` or `scheduled_date_start` (new `_dayroute_trigger_fields()` hook); any other field write returns early via `super().write(vals)`, untouched.
- Each record gets its own copy of `vals` and its own `super(FSMOrder, rec).write(rec_vals)` call, so a batch write can no longer cross-contaminate records.
- Unassigning an order (both `person_id` and `scheduled_date_start` cleared — e.g. by an auto-reschedule cron) now clears `dayroute_id` too, instead of leaving it pointing at a dayroute it no longer belongs to.
- The now-emptied dayroutes are deleted *after* the per-record writes have actually happened.
- `_get_dayroute_domain()` and `check_capacity()` both treat `max_order = 0` as "no limit", consistent with the field's default and help text: `'|', ('max_order', '=', 0), ('order_remaining', '>', 0)`.
- The cleanup itself is now a small overridable pair, `_is_removable()` / `_unlink_removable()`, instead of being inlined at the two call sites — so a downstream module that hangs other content off a dayroute (e.g. helper technicians' own time slots, not just `order_ids`) can veto the deletion without having to reimplement or monkeypatch the cleanup.

8 new tests in `fieldservice_route/tests/test_fsm_order.py` reproduce each defect against `HEAD` (6 fail/error without the fix) and pass with it; the existing test suite plus `fieldservice_sale_stock_route`, `fieldservice_route_availability` and `fieldservice_calendar` (all of which touch dayroutes) were run together against the change with no regressions.

## Known limitations (out of scope for this PR)

- `_get_dayroute_values()`'s `.date()` call on `scheduled_date_start` doesn't account for the user's timezone.
- `date_start_planned` is hardcoded to 8:00 (there's already a `# TODO: Use the worker timezone and working schedule` upstream).
- `_default_team_id()` raises if no `fsm.team` exists yet.

## Semantics change to flag for review

Before this fix, `max_order = 0` effectively meant "reject everything after the first write" (an unusable state, given it's the field's own default). After this fix it means "no limit", matching the help text. If reviewers would rather keep an explicit capacity requirement, the domain could instead read `('order_remaining', '>', 0)` gated by `max_order > 0` only — happy to adjust.

### DIFF
--- a/fieldservice_route/README.rst
+++ b/fieldservice_route/README.rst
@@ -45,23 +45,24 @@ Configuration
 
 To use this module, you need to:
 
--  Go to Field Service > Configuration > Stages
--  Define the workflow of your routes by creating new stages whose type
-   is "Route"
--  Go to Field Service > Master Data > Routes
--  Create your routes by setting their name and selecting their
-   territory
+- Go to Field Service > Configuration > Stages
+- Define the workflow of your routes by creating new stages whose type
+  is "Route"
+- Go to Field Service > Master Data > Routes
+- Create your routes by setting their name and selecting their territory
+- Set "Maximum Orders" on a route to cap how many orders a day route for
+  it can hold; leave it at its default of 0 for no limit
 
 Usage
 =====
 
 To use this module, you need to:
 
--  Go to Field Service
--  Create or select an order
--  Assign it to a worker and schedule it
--  Go to Field Service > Dashboard > Day Routes. A new record has been
-   created.
+- Go to Field Service
+- Create or select an order
+- Assign it to a worker and schedule it
+- Go to Field Service > Dashboard > Day Routes. A new record has been
+  created.
 
 Bug Tracker
 ===========
@@ -84,20 +85,20 @@ Authors
 Contributors
 ------------
 
--  Maxime Chambreuil <mchambreuil@opensourceintegrators.com>
--  Nikul Chaudhary <nikul.chaudhary.serpentcs@gmail.com>
--  Freni Patel <fpatel@opensourceintegrators.com>
--  `Tecnativa <https://www.tecnativa.com>`__:
+- Maxime Chambreuil <mchambreuil@opensourceintegrators.com>
+- Nikul Chaudhary <nikul.chaudhary.serpentcs@gmail.com>
+- Freni Patel <fpatel@opensourceintegrators.com>
+- `Tecnativa <https://www.tecnativa.com>`__:
 
-   -  Víctor Martínez
+  - Víctor Martínez
 
 Other credits
 -------------
 
 The development of this module has been financially supported by:
 
--  Open Source Integrators <https://opensourceintegrators.com>
--  Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
+- Open Source Integrators <https://opensourceintegrators.com>
+- Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
 
 Maintainers
 -----------
--- a/fieldservice_route/__manifest__.py
+++ b/fieldservice_route/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Field Service Route",
     "summary": "Organize the routes of each day.",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.0.1",
     "category": "Field Service",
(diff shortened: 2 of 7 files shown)


ppyczko (approved):
LGTM!

miquelpascual (approved):
LGTM!

BernatObrador (approved):
LGTM

ppyczko:
Hi @max3903 @brian10048 can you check if this can be merged? Thanks!

max3903 (approved):
## Review summary

**Verdict:** approve
**CI:** green (pre-commit, tests Odoo/OCB, codecov patch/project, runboat/build, unreleased deps)
**Module(s):** `fieldservice_route`

### Suggestions
- [ ] PR description: add a short `## Test scenario` with numbered UI/API steps (the automated tests cover the defects well; a concrete manual scenario still helps reviewers).
- [ ] Consider whether `fsm_route_id` (or location-driven route changes) should join `_dayroute_trigger_fields()` so a route-only write rebinds `dayroute_id` (optional; current scope correctly fixes person/date/capacity duplicates).

### Checked
- [x] OCA conventions / manifest (`development_status`, `maintainers`, AGPL-3)
- [x] Coding standards for 18.0
- [x] CI (`gh pr checks`)
- [x] Bugbot (no findings)
- [x] Security Review (no medium+)
- [x] Performance (manual — per-record `write` is intentional to avoid vals cross-contamination; acceptable for FSM order batches)

### Notes
Solid fix for the production duplicate/orphan dayroute issues: gate dayroute logic on trigger fields, copy `vals` per record, treat `max_order=0` as unlimited, and unlink empty dayroutes after the move. Regression tests look thorough.

PR #1612, opened 2026-08-22
[IMP] base_territory: Menus, extra data state mgmt 

1AbdulkarimMousa:
UI to manage territories, regions, districts, and hardening the partner form to prevent geo-mismatches


base_territory/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Base Territory",
     "summary": "This module allows you to define territories, branches,"
     " districts and regions to be used for Field Service operations or Sales.",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.1.0",
```
amh-mw: Versions are incremented by the merge process; not manually.

base_territory/static/src/js/district_polygon_editor.esm.js
```diff
@@ -0,0 +1,128 @@
...
+
+import {loadJS} from "@web/core/assets";
+import {rpc} from "@web/core/network/rpc";
+
+let mapsPromise = null;
+
+async function loadGoogleMaps() {
+    if (!mapsPromise) {
+        mapsPromise = rpc("/web/dataset/call_kw", {
+            model: "ir.config_parameter",
```
amh-mw: How can this work for normal users? Per https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/security/ir.model.access.csv#L118, only super user has any access to the `ir.config_parameter` model. Consider controller route like https://github.com/odoo/odoo/blob/f9d53aeaa9a5442587492549bc6f1c4314b4ddd0/addons/website/controllers/main.py#L999-L1003.

base_territory/models/res_district.py
```diff
@@ -12,3 +12,36 @@ class ResDistrict(models.Model):
...
+    polygon_ids = fields.One2many(
+        "res.district.polygon.point", "district_id", string="Polygon Points"
+    )
+
+    @api.model
+    def find_by_coordinates(self, lat, lng):
+        if lat is None or lng is None:
+            return self.env["res.district"]
+        districts = self.search([("polygon_ids", "!=", False)])
+        for district in districts:
```
amh-mw: Doesn't postgres support PostGIS for geospatial?

amh-mw (changes requested):
Code review only: `base_territory`

Per comments in Discord, commits should to be updated per https://github.com/OCA/.github/blob/master/AI_POLICY.md

base_territory/models/res_district.py
1AbdulkarimMousa: PostGIS supports this, but most hosting providers don't support the postgres extension

1AbdulkarimMousa: Its better to live within /OCA/geospatial, PostGIS is the standard, so should this be moved into a separate module ?

max3903:
@1AbdulkarimMousa You should read https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst. You will learn that you should make 1 pull request per module.

1AbdulkarimMousa:
I'm building a better hierarchy of my edits, this PR will be only of base_territory then will build my extra modules around it

1AbdulkarimMousa:
image

Also, selecting a district will auto-select the parent region → state → country fields. The same logic applies when selecting a region or state, which double-ensures that geo mismatches will never happen.

1AbdulkarimMousa:
Re-trigger CI against current head ebcb10c5b (previous run tested a stale snapshot showing code 'TC' that no longer exists in the test file).

1AbdulkarimMousa:
image
image
image


the geo mismatches are resolved automatically

PR #1613, opened 2026-08-26
[FIX] fieldservice: Quick search finds locations with no reference

SirPyTech:
**Steps**:
1. Create two locations:
    - one only has Name "Test ref location" and no Reference
    - one has both Name and Reference equal to "Test ref location"
2. In an order, click on the Plant field: check that both locations are shown image
3. Type `ref` in the field

**Current behavior**:
Only the location that has both Name and Reference is shown: image

**Expected behavior**:
Both locations are still shown.
This is consistent with searching `ref` in the Locations menu: image
because the filter is using both `name` and `ref`:
https://github.com/OCA/field-service/blob/868ecdeab46e0e383c424e317c29046925086704/fieldservice/views/fsm_location.xml#L288-L292

### DIFF
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -10,6 +10,12 @@ class FSMLocation(models.Model):
     _inherits = {"res.partner": "partner_id"}
     _inherit = ["mail.thread", "mail.activity.mixin", "fsm.model.mixin"]
     _description = "Field Service Location"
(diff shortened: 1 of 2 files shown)


max3903 (approved):

HekkiMelody (approved):
Code and functional review, LGTM

SirPyTech:
Thanks! :pray:

PR #1466, opened 2025-11-05
[17.0][FIX] fieldservice_sale: Correctly compute template on FSM order creation

BernatObrador:
Before this change, when using a single product with FSM tracking set to 'sale', the FSM order was not auto-completed with its template. Now the template is properly assigned during order creation.

PR #1614, opened 2026-08-27
[16.0][FIX] fieldservice_sale: Correctly compute template on FSM order creation

LorenzoC0:
From #1466 :

> @BernatObrador: Before this change, when using a single product with FSM tracking set to 'sale', the FSM order was not auto-completed with its template. Now the template is properly assigned during order creation.

This is also needed for 16.0, so I'm cherry-picking the commit from that PR.

### DIFF
--- a/fieldservice_sale/models/sale_order.py
+++ b/fieldservice_sale/models/sale_order.py
@@ -79,6 +79,10 @@ class SaleOrder(models.Model):
         self.ensure_one()
         template_id = kwargs.get("template_id", False)
         template_ids = kwargs.get("template_ids", [template_id])
+        # If only 1 template_ids is given, use it as template_id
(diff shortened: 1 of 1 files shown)


max3903 (approved):
LGTM. Clean cherry-pick of the merged 17.0 fix (#1466).

Both changes are needed: setting `template_id` from a single `template_ids` entry, and calling `_onchange_template_id()` after create so template `type`/`team_id` (and related fields) are applied — onchange does not run on programmatic create.

CI is green. Optional follow-up: forward-port to 18.0 (still missing) and ensure 19.0 also invokes `_onchange_template_id()` after sale-level FSM order create.

mrcast (approved):
@LorenzoC0 functional test: ok!

PR #1617, opened 2026-09-08
[FIX] fieldservice: Quick search finds locations with no reference

SirPyTech:
Forward port of https://github.com/OCA/field-service/pull/1613.

Here there was no issue because the model is always searching on `complete_name`, but `search_on_complete_name` was ignored, so I aligned the code to the linked PR.

### DIFF
--- a/fieldservice/models/fsm_location.py
+++ b/fieldservice/models/fsm_location.py
@@ -10,8 +10,15 @@ class FSMLocation(models.Model):
     _inherits = {"res.partner": "partner_id"}
     _inherit = ["mail.thread", "mail.activity.mixin", "fsm.model.mixin"]
     _description = "Field Service Location"
+    # Is overridden by search_on_complete_name (see `name_search`)
+    _rec_names_search = [
+        "email",
+        "name",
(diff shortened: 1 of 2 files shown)


HekkiMelody (approved):

max3903 (approved):
Forward-port of #1613 looks correct for 17.0.

**What works**
- Honors `company.search_on_complete_name` (previously ignored whenever `name` was set).
- Falls back to `super().name_search` with `_rec_names_search` including `name`/`ref` plus the 17.0 extras (`email`, `partner_id.vat`).
- `_rec_name = "complete_name"` aligns display with 16.0’s `name_get()` (17.0 had no `name_get` override).
- Test covers the default (non-hierarchy) path; Odoo/OCB CI green.

**Nits (non-blocking)**
- `codecov/patch` fails because the `search_on_complete_name=True` branch is untested; a small toggle+assert would close that.
- As the author notes, the exact 16.0 “ref-only” bug was not present on 17.0; this is still a useful alignment and adds `ref` to quick search.

PR #1609, opened 2026-08-19
[FIX] fieldservice_geoengine: Batch creation

SirPyTech:
When attempting to create multiple locations at once, the following is raised:
```
[...]
  File "/opt/odoo/auto/addons/fieldservice_geoengine/tests/test_fsm_location.py", line 172, in test_multi_create
    locations = self.FSMLocation.create(
  File "<decorator-gen-161>", line 2, in create
  File "/opt/odoo/custom/src/odoo/odoo/api.py", line 431, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/auto/addons/fieldservice_geoengine/models/fsm_location.py", line 19, in create
    if not res.partner_latitude or not res.partner_longitude:
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1152, in __get__
    record.ensure_one()
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5222, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: fsm.location(15, 16)
```

PR #1618, opened 2026-09-08
[FIX] fieldservice_geoengine: Batch creation

SirPyTech:
Forward porting https://github.com/OCA/field-service/pull/1609.


HekkiMelody (approved):
LGTM

max3903 (approved):
Forward-port of #1609 looks correct.

`create` is `model_create_multi` but the old code accessed `partner_latitude`/`partner_longitude` on the whole recordset (`ensure_one` crash on batch). Filtering non-localized records and calling `geo_localize()` on the subset is the right fix; `geo_localize` already handles multi via `mapped("partner_id")`.

Test covers the crash path; CI green.

PR #1453, opened 2025-10-08
[16.0][IMP] fieldservice_geoengine: Fix vector Layer related Errors

anusriNPS:
- [ ] Depends on base_geoengine PR [#417](https://github.com/OCA/geospatial/pull/417#top)

Domain for "attribute_field_id" is filtered based on "classification" field selection in vector layer, which helps to avoid JS error seen as "fsm.order" and "fsm.location" related fields are not validated based on "NUM_ATT" types.[ base_geoengine_validation](https://github.com/OCA/geospatial/blob/16.0/base_geoengine/models/geo_vector_layer.py#L122).

   Notifying user to define selected field from attribute_field_id of supported type in geoengine view which avoids below JS error observed:
   Screenshot from 2025-10-08 17-09-18

   Notifying users when a field of supported type from attribute_field_id is not suitable to use for classification type "custom" which avoids below observed JS Error:
   image
   
  As  fieldservice_geoengine omits validation for custom classification related to "type" check on field selected in "attribute_field_id", when user selects fields of type "date" with "custom", below JS error observed:
   image

PR #1621, opened 2026-09-16
[17.0][FIX] fieldservice_geoengine: Fix vector Layer related Errors

anusriNPS:
FW #1453 
  
  Depends on geospatial PR [#481](https://github.com/OCA/geospatial/pull/481)


max3903 (changes requested):
Thanks for the forward-port of #1453. Please address the following before merge:

### Blockers
1. **Dependency on geospatial** — This needs [OCA/geospatial#481](https://github.com/OCA/geospatial/pull/481) merged first. Please add label `blocked by dependency` and document the dependency in the PR body (same pattern as #1453 ↔ geospatial#417).
2. **Incomplete base FW?** — geospatial#481 currently only adds JS `checkAttributeFieldUsage`. Confirm that `attribute_field_id_domain` exists on 17.0 `geoengine.vector.layer`; otherwise `_compute_attribute_field_id_domain` here is a no-op and the domain half of the fix is missing.
3. **Wrong `_t` import** — `import {_t} from "web.core"` is incorrect for this OWL ESM module. Align with base_geoengine and use `@web/core/l10n/translation`.

### Other issues
- **`it.po` header regression** — restores wipe Language / Plural-Forms / Weblate translator metadata. Keep existing headers; only add the new msgids.
- **Silent `try/catch`** around label styling swallows errors — please log or notify instead of bare `return`.
- **Fragile heuristics** — `vals.some((item) => item === 0)` and `"0 - 0"` can false-positive/false-negative; consider tighter checks or document the limitation.
- **Tests** — please add coverage for `_compute_attribute_field_id_domain` (and keep JS behavior covered by the dependency PR where appropriate).
- Expand the PR body beyond `FW #1453` with a short summary + test scenario + dependency note.

